PR #41, opened 2016-10-28
Feature/cc3200 build doc add device

DigTheDoug:
Including @philskaroulis structure changes.
The larger changes are the new sections around setting up your device in CPM and what happens at the end if it's successful.


DellaBitta:
I will tackle most of these changes olgierd
the last one we should do after the document merge.

PR #45, opened 2016-11-09
Feature/improve logging system

olgierdh:
[DESCRIPTION]

It fixes the functionality of printing library debug logs using any defined function that is printf compatible. By default when there is no XI_DEBUG_PRINTF defined it uses just printf. When XI_DEBUG_PRINTF=PrintLikeFunctionName the library will use the PrintLikeFunctionName in order to printf debug logs. 

It also adds an information about the time to the logs. Each log entry now has a timestamp at the beginning. 

In addition it prevents the debug logger from displaying whole file path of __FILE__ it just displays the file name which shortens the logs and makes it more readable.

Example of new output:

```
[1478658951584][xi_mqtt_logic_layer.c:290 (xi_mqtt_logic_layer_pull)] [m.id[1] m.type[9]] received msg
[1478658951584][xi_mqtt_logic_layer_subscribe_command.h:145 (do_mqtt_subscribe)] [m.id[1]]subscribe suback received
[1478658951584][xi_control_topic_layer.c:148 (xi_on_control_message)] Subscription to control topic successfull with QoS 1
[1478658951584][xi_user_sub_call_wrapper.c:101 (xi_user_sub_call_wrapper)] user callback returned with state: , (0)
[1478658951584][xi_mqtt_logic_layer_connection_helpers.h:54 (do_shutdown)] message memory filled with disconnect data
```

[REVIEWERS]
@DellaBitta

[QA]

Tested on CC3200 using TI ARM compiler and UART debug logging.

[RELEASE NOTES]

None?


src/libxively/xi_debug.c
```diff
@@ -37,3 +38,29 @@ void xi_debug_data_logger_impl( const char* msg, const xi_data_desc_t* data_desc
 
     xi_free( tmp );
 }
+
+/** Purpose of this function is to cut the file name */
```
DellaBitta: Trims the absolute file path to just the file name.

olgierdh: Roger

src/libxively/xi_debug.c
```diff
@@ -37,3 +38,29 @@ void xi_debug_data_logger_impl( const char* msg, const xi_data_desc_t* data_desc
 
     xi_free( tmp );
 }
+
+/** Trims the absolute file path to just the file name */
+const char* xi_debug_dont_print_the_path( const char* msg )
```
atigyi: I was wondering about compile time solutions. Searched for another macro but there is not __ONLY_FILENAME__ or such. But they claim
gcc file.c
gcc path/to/source/file/file.c
http://stackoverflow.com/questions/8487986/file-macro-shows-full-path
sets the __FILE__ macro. We cannot use this but
http://stackoverflow.com/questions/7398310/how-to-get-only-file-name-in-preprocessor
says a compiler flag defined as this:
-DXI_BASE_FILENAME=\"$(<F)\"
could provide you the only the filename. Which you could use in the print functions instead of __FILE__

What about this compile time solution vs the runtime path parsing?

atigyi:
As we see here Olgierd selects the debug output locally here in the BSP wolf solution. Without including xi_debug.h.
I wouldn't define this for all BSP .c implementations. Although cannot include xi_debug.h because it refers to libxively internals as well. Another question here is whether we would like to "teach" or load the porting developer with our libxively internal debug mechanism?
I ask these question because you - understandably and validly - pointed on my bare printf solutions. Which printfs are there because I don't know the answers on the above questions.

atigyi:
As an "ideal" solution - I can imagine now - would be to separate xi_debug.h into two parts:
1 - libxively internal defines
2 - porting developer compatible defines which are clear and understandable by a porting developer and any BSP .c file can include it.

src/libxively/xi_debug.c
olgierdh: That's interesting, I've never checked it - I don't know if this will work with majority of compilers - if we can provide such solution then I'm for it. The solution I've implemented does not depend on the compiler features... so we can explore compiler based solution.

olgierdh:
In terms of using the debug output in BSP implementations I have one idea that should do the job - I will prepare proposal for it - essentially it's about moving logging debug macros into the /include/xi_debug_logging.h then make it includable into xi_debug.h and in any BSP implementation.

src/libxively/xi_debug.c
atigyi: I don't think any special compiler flag is required. It is a macro defined by -D . The make system fills it with proper content: filename

PR #46, opened 2016-11-10
Feature/enumerate error codes

DellaBitta:
[ Description ]
Added comments to the error code enumeration (xi_error.h) and the corresponding string table (xi_err.c) to quickly correlate an error number to an enumeration / error name.  

Added some missing strings to the string error table to represent errors that had been recently added to the error code enumeration.

[ Reviewer ]
@olgierdh 

[ QA ]
Build with error strings and without.

[ Release Notes ]
Added error code numbers to comments in xi_error.h to easily map error codes integer values to their corresponding enumerated type.


DellaBitta:
yes it was!

PR #48, opened 2016-11-14
Fix/make the layers work with large buffers

olgierdh:
[DESCRIPTION] 

Fix for the problem with not checking if whole data has been read from the TLS's library read buffer and from local raw buffers list.

[REVIEWERS]

@atigyi 
@DellaBitta 

[QA]

New functional tests had been added in order to test the solution. All other tests pasess.

[RELEASE NOTES]

Fixed bug where MQTT messages stacked together where not received when the receive buffer size was greater then the MQTT msg size.


src/bsp/tls/wolfssl/xi_bsp_tls_wolfssl.c
```diff
@@ -420,3 +420,14 @@ xi_bsp_tls_state_t xi_bsp_tls_cleanup( xi_bsp_tls_context_t** tls_context )
...
+
+int xi_bsp_tls_pending( xi_bsp_tls_context_t* tls_context )
+{
+    wolfssl_debug_format( "[ %s ]", __FUNCTION__ );
+
+    /* get back the wolfssl_tls_context */
+    wolfssl_tls_context_t* wolfssl_tls_context = tls_context;
+    assert( NULL != wolfssl_tls_context->obj );
+
+    return CyaSSL_pending( wolfssl_tls_context->obj );
```
atigyi: wolfssl_pending?

src/libxively/tls/xi_tls_layer.c
```diff
@@ -333,6 +326,13 @@ static xi_state_t recv_handler( void* context, void* data, xi_state_t in_out_sta
     xi_data_desc_t* ret_buffer = layer_data->decoded_buffer;
     layer_data->decoded_buffer = NULL;
 
+    if ( NULL != layer_data->raw_buffer ||
+         xi_bsp_tls_pending( layer_data->tls_context ) > 0 )
```
atigyi: What is the benefit of using the TLS library's pending solution vs. libxively's internal pending "solution"?
I ask this because this raises complexity in BSP TLS module which hardens porting, increases BSP TLS API functions from 5 to 6.

src/bsp/tls/mbedtls/xi_bsp_tls_mbedtls.c
```diff
@@ -335,6 +335,15 @@ xi_bsp_tls_state_t xi_bsp_tls_write( xi_bsp_tls_context_t* tls_context,
     return XI_BSP_TLS_STATE_WRITE_ERROR;
 }
 
+int xi_bsp_tls_pending( xi_bsp_tls_context_t* tls_context )
+{
+    assert( NULL != tls_context );
+
+    mbedtls_tls_context_t* mbedtls_tls_context = tls_context;
+
+    return ( int )mbedtls_ssl_get_bytes_avail( &mbedtls_tls_context->ssl );
```
atigyi: 👍 super you covered both TLS BSPs.

src/libxively/mqtt/codec/xi_mqtt_codec_layer.c
```diff
@@ -277,10 +277,16 @@ xi_state_t xi_mqtt_codec_layer_pull( void* context, void* data, xi_state_t in_ou
         layer_data->local_state =
             xi_mqtt_parser_execute( &layer_data->parser, layer_data->msg, data_desc );
 
+        xi_debug_format( "local_state = %d", ( int )layer_data->local_state );
+
+        if ( layer_data->local_state == XI_STATE_WANT_READ )
+        {
+            xi_free_desc( &data_desc );
```
olgierdh: Yes essentially if the layer_data->local_state == XI_STATE_WANT_READ whole memory from the buffer has been drained by the parser so it is safe to release that. I agree that I should put here some assertions in order to make it super secured. Tested with both too big and too small buffers.

src/libxively/tls/xi_tls_layer.c
```diff
@@ -275,18 +280,6 @@ static xi_state_t recv_handler( void* context, void* data, xi_state_t in_out_sta
         return XI_STATE_OK;
     }
 
-    /* if we suppose to reuse the buffer */
-    if ( XI_STATE_WANT_READ == in_out_state && NULL != data_desc )
-    {
```
olgierdh: Ok, so the story begins with the xi_mqtt_codec_layer - the xi_mqtt_codec_layer, before that PR change, could call the xi_tls_layer::pull function in order to notify it that it requires more data. In case if the parser couldn't finalise reading a message. After I added your patch plus checks for pending data to xi_tls_layer xi_mqtt_codec_layer does no longer need to do this, becuase the xi_tls_layer is self aware of the situation. In fact that xi_mqtt_codec_layer demand of more data was an overuse because of the lack of other solution that's why it has been taken out. So by taking that change out - we do no longer need a to check for this condition within the xi_tls_layer that's why we can remove that code.

src/libxively/tls/xi_tls_layer.c
olgierdh: Yes, you are correct - so it does rise the complexity - but is the only way to guarantee to process whole accessible data to read from both libxively and tls libraries. So there is one case where the memory from the tls read buffer may stay untouched until the next message triggers it.

src/bsp/tls/mbedtls/xi_bsp_tls_mbedtls.c
olgierdh: :)

include/bsp/xi_bsp_tls.h
```diff
@@ -180,6 +180,20 @@ xi_bsp_tls_state_t xi_bsp_tls_read( xi_bsp_tls_context_t* tls_context,
                                     uint8_t* data_ptr,
                                     size_t data_size,
                                     int* bytes_read );
+
+/**
+ * @function
+ * @brief Implements the TLS pending.
+ *
+ * Implementation of this function should call TLSs library pending function. It should
+ * return number of bytes which are buffered and available in the TLS library to be read
```
DellaBitta: It should return the the number of bytes that are available to be read by the tls library's read function.

src/bsp/tls/wolfssl/xi_bsp_tls_wolfssl.c
DellaBitta: yes we should migrate toward using WolfSSL functions.

olgierdh: Yes, true.

include/bsp/xi_bsp_tls.h
olgierdh: Roger roger, thanks!

src/libxively/mqtt/codec/xi_mqtt_codec_layer.c
atigyi: ok, tests MUST come here to cover these. The story should include these integration tests as well!

src/libxively/tls/xi_tls_layer.c
atigyi: aha, so cross-question:
so the TLS lib's _pending_ function (e.g. wolfssl_pending) returns 0< in two cases:
- there are further TLS packages unprocessed (this is the case you found and my bugfix primarily targets)
- the current TLS package is not fully read

atigyi: here my question is regarding the 1st:
As I see in xi_tls_layer.c the layer_data->raw_buffer is filled when package arrived on socket. How will TLS library know that this layer_data->raw_buffer buffer contains more without interacting with it?
layer_data->raw_buffer is a list. Suppose there are 2 elements in the list. TLS layer reads the first element fully. And this element is either one or more full TLS packages. Which means TLS package does not span over the border of the two list elements. How will TLS library know that the raw_buffer contains another element?

src/libxively/mqtt/codec/xi_mqtt_codec_layer.c
olgierdh: Yes, tests will come :) I just wanted to start the discussion on this solution.

src/libxively/tls/xi_tls_layer.c
olgierdh: Simply - the layer_data->raw_buffer is tested for existance -  I'm checking if the layer_data->raw_buffer exists != NULL - this is sufficient because the tls_callback_read function copies data from the raw_buffer and deletes it when it's emptied - so if there is a single element on that list it means there is data to read. Does it answer your questions ?

src/libxively/tls/xi_tls_layer.c
olgierdh: Whenever the data is left within the TLS read buffer and it does not reside in any external ( to the TLS's library buffer ) read_buffer.

PR #52, opened 2016-11-16
Carriage Returns Removed from Output

DellaBitta:
[ Description ]
There were places where some of our output used \n\r while other places used just \n.  This small PR makes our usage of new lines more consistent by removing uses of \r.

[JIRA]
XCL-2225
Make examples printf usage consistent

[ Reviewer ]
@atigyi 

[ QA ]
Built the library and tutorials to ensure there were no syntax errors.

[ Release Notes ]
Removed uses of \r from some debug output statements to be more consistent, and to remove some formatting problems on some platforms.  We now always simply use \n.


src/examples/internal/mqtt_logic_example_tls_bsp.c
```diff
@@ -29,9 +29,9 @@ static uint8_t connected    = 0;
```
atigyi: I don't think we need this file at all.

DellaBitta: ok, I'll ditch it!

atigyi:
👍 approved

PR #56, opened 2016-11-24
Remove link to wolfssl library from xively demo

olgierdh:
[DESCRIPTION]
We don't need and we don't want wolfssl library to be linked against our xively demo example for CC3200 since the xively demo is using on board TLS implementation. 

[REVIEWERS]
@DellaBitta 

[QA]
Tested building of xively demo without compiled wolfssl library.

[RELEASE NOTES]
N/A


atigyi:
👍 approved

PR #66, opened 2016-12-06
Feature/add clang fuzz testing and static analyzis

olgierdh:
[DESCRIPTION]

Added support for fuzz tests for the xively-client-c library. Fuzz tests are integrated using libFuzzer a library from LLVM. 

In addition a special target for clang-tidy has been added so it's easy to run the statical analysis of the library code.  

Fixed problem found within the xi_mqtt_parser where the static variables were used to track the parser coroutine state. It has prevented xi_mqtt_parser to be tested using libFuzzer.  

[JIRA]

https://jira.3amlabs.net/browse/XCL-2667

[REVIEWERS]
@DellaBitta

[QA]

All remaining tests passes. 

New kind of tests has been added - fuzz tests. This will increase the reliability of components such as parsers and other network components of the library that have to handle data received over the wire. 

[RELEASE NOTES]

From now on the library supports fuzz tests. For more details refer to the doc/FuzzTestsProgrammerManual.md.


doc/FuzzTestsProgrammerManual.md
```diff
@@ -0,0 +1,47 @@
+# Fuzz Testing on xively-client-c IoT library
```
atigyi: two things regarding the file itself:
1 - naming, isn't this the first CamelCase named file in the repo? What about fuzz_tests_programmer_manual.md, isn't his more libxively style?
2 - location. I don't think this programming manual is for the same interest level visitors, so I wouldn't put this on the same level as porting guide and user guide. I'd put this under doc/dev ?
And I've found another camel cased filename doc/res

make/mt-config/tests/mt-native
```diff
@@ -6,3 +6,4 @@
 #XI_RUN_UTESTS = $(XI_UTESTS) -l0 --terse
 XI_RUN_UTESTS := (cd $(dir $(XI_UTESTS)) && LD_LIBRARY_PATH=$(dir $(XI)):$$LD_LIBRARY_PATH exec $(XI_UTESTS) -l0)
 XI_RUN_ITESTS := (cd $(dir $(XI_ITESTS)) && LD_LIBRARY_PATH=$(dir $(XI)):$$LD_LIBRARY_PATH exec $(XI_ITESTS))
+XI_RUN_FTEST = (cd $(XI_FUZZ_TESTS_BINDIR) && $(1) $(XI_FUZZ_TESTS_CORPUS_DIR)/$(notdir $(1))/ -max_total_time=10 -max_len=256);
```
olgierdh: We might want to parameterize fuzz test run settings. Next step might be also parallelization of test run so that bigger domain is checked.

.travis.yml
```diff
@@ -1,4 +1,4 @@
-sudo: false
+sudo: true
```
atigyi: is sudo a need? seems it significantly increases boot time of travis:
https://docs.travis-ci.com/user/ci-environment/
Also this page says "sudo: required" instead of "sudo: true", but maybe I looking at wrong page.
The point here is that using sudo seems to have effect, is it needed really?

olgierdh: Yes I needed sudo for some moment in time ( external scripts required it for some reason ) now I'll switch it back and test if it works without sudo. Good catch, thanks.

Makefile
```diff
@@ -116,7 +117,6 @@ $(XI_OBJDIR)/tests/tools/xi_libxively_driver/%.o : $(LIBXIVELY)/src/tests/tools/
 	@-mkdir -p $(dir $@)
 	$(info [$(CC)] $@)
 	$(MD) $(CC) $(XI_CONFIG_FLAGS) $(XI_COMPILER_FLAGS) $(XI_INCLUDE_FLAGS) $(XI_TEST_TOOLS_INCLUDE_FLAGS) -c $< -o $@
-	@$(CC) $(XI_CONFIG_FLAGS) $(XI_COMPILER_FLAGS) $(XI_INCLUDE_FLAGS) $(XI_TEST_TOOLS_INCLUDE_FLAGS) -MM $< -MT $@ -MF $(@:.o=.d)
```
atigyi: seems the libxively_driver's .o file generation is turned off by this. Are the functional tests still working?

Makefile
```diff
@@ -183,11 +183,37 @@ ifneq ($(XI_CONST_PLATFORM_CURRENT),$(XI_CONST_PLATFORM_ARM))
 
 $(XI_ITESTS): $(XI) $(CMOCKA_LIBRARY_DEPS) $(XI_ITEST_OBJS)
 	$(info [$(CC)] $@)
-	$(MD) $(CC) $(XI_ITEST_OBJS) -L$(XI_BINDIR) $(XI_LIB_FLAGS) $(CMOCKA_LIBRARY) -o $@
+	$(MD) $(CC) $(XI_ITEST_OBJS) $(XI_ITESTS_CFLAGS) -L$(XI_BINDIR) $(XI_LIB_FLAGS) $(CMOCKA_LIBRARY) -o $@
 	$(XI_RUN_ITESTS)
-
 endif
 
+#$(XI_OBJDIR)/%.cpp.o : $(LIBXIVELY)/src/%.cpp $(XI_BUILD_PRECONDITIONS)
```
atigyi: is this commented-out block still needed?

olgierdh: it is not, thanks.

Makefile
olgierdh: Hmm I must've made a mistake here, thanks for catching it.

make/mt-config/mt-config
```diff
@@ -56,7 +56,12 @@ ifneq (,$(findstring debug,$(TARGET)))
 	XI_DEBUG_EXTRA_INFO ?= 1
 endif
 
-XI_COMPILER_FLAGS += -Wall -Werror
+# Settings that will work only on linux and only against clang-4.0 and greater
+ifneq (,$(findstring fuzz_test,$(CONFIG)))
+    XI_CONFIG_FLAGS += -fsanitize=address -fomit-frame-pointer -fsanitize-coverage=trace-pc-guard -g
+endif
+
+XI_COMPILER_FLAGS += -Wall
```
atigyi: I am not certaint of -Wall, there are some compilers where some were intentionally turned off e.g. mt-arm-linux: XI_COMPILER_FLAGS += -Wno-format. Another reason for holding back this -Wall is the false positives of Travis. So here we cannot rely on Travis' greens.

make/mt-config/mt-presets.mk
```diff
@@ -6,11 +6,12 @@
 include make/mt-config/mt-target-platform.mk
 
 # CONFIG presets
-CONFIG_POSIX_MAX                =bsp_posix-posix_fs-posix_platform-tls_bsp-senml-control_topic-memory_limiter
+CONFIG_POSIX_MAX                =bsp_posix-posix_fs-posix_platform-tls_bsp-control_topic-memory_limiter
 CONFIG_POSIX_MAX_THREADING      =bsp_posix-posix_fs-posix_platform-tls_bsp-senml-control_topic-threading-memory_limiter
 CONFIG_POSIX_MID                =bsp_posix-posix_fs-posix_platform-tls_bsp-senml-control_topic
 CONFIG_POSIX_MID_UNSECURE       =bsp_posix-posix_fs-posix_platform-senml-control_topic
 CONFIG_POSIX_MIN                =bsp_posix-posix_fs-posix_platform-tls_bsp
+CONFIG_POSIX_MIN_UNSECURE       =bsp_posix-posix_fs-posix_platform
```
atigyi: I think for not raising complexity more higher please remove the CONFIG_POSIX_MID_UNSECURE

src/libxively/mqtt/codec/xi_mqtt_parser.c
```diff
@@ -489,6 +493,7 @@ xi_state_t xi_mqtt_parser_execute( xi_mqtt_parser_t* parser,
     {
         xi_debug_logger( "XI_MQTT_ERROR_PARSER_INVALID_MESSAGE_ID" );
         parser->error = XI_MQTT_ERROR_PARSER_INVALID_MESSAGE_ID;
+        XI_CR_RESET( parser->read_cs );
```
atigyi: so this cs->parser->read_cs is a "bugfix" to support fuzzer, right?

make/mt-config/mt-config
olgierdh: But I didn't enable it, it was already enabled I've just moved the whole section. But in terms of -Wall I would leave it as it increases the quality of compiler tests - and this is the first line of defence. I would like to enable -pedantic but maybe some other time :)

src/tests/fuzztests/xi_fuzztest_mqtt_layer.cpp
```diff
@@ -0,0 +1,10 @@
+#include <memory.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <string.h>
+
+extern "C" int LLVMFuzzerTestOneInput( const uint8_t* data, size_t size )
```
atigyi: again CamelCase vs. underscore_separated_naming inside and in filename

make/mt-config/mt-presets.mk
```diff
@@ -6,11 +6,12 @@
 include make/mt-config/mt-target-platform.mk
 
 # CONFIG presets
-CONFIG_POSIX_MAX                =bsp_posix-posix_fs-posix_platform-tls_bsp-senml-control_topic-memory_limiter
+CONFIG_POSIX_MAX                =bsp_posix-posix_fs-posix_platform-tls_bsp-control_topic-memory_limiter
```
olgierdh: Sorry I've mixed some in the middle of testing changes - fixing.

src/tests/fuzztests/xi_fuzztest_mqtt_layer.cpp
atigyi: ah is this a function which has to be implemented and executed by the fuzzer right? so you cannot determine the naming, then it is ok, sorry.

src/libxively/mqtt/codec/xi_mqtt_parser.c
olgierdh: Sort of - the problem with local static variable that kept the coroutine state was that in between fuzz tests calls the static variable couldn't be zeroed - so the elements of the parser were "continuing" execution from the remembered state. This could also be a potential issue with restarting the the xively client library while still receiving a msg.

src/tests/fuzztests/xi_fuzztest_mqtt_layer.cpp
olgierdh: yes, this is the one that is expected and external.

doc/FuzzTestsProgrammerManual.md
olgierdh: Ok, thanks for the suggestions.

make/mt-config/tests/mt-native
atigyi: how many cores Travis machines have? I guess 2. So the it can be accelerated two-times at most currently. Isn't this true?

make/mt-config/mt-config
atigyi: true, ok :+1:

make/mt-config/tests/mt-native
olgierdh: I was more thinking of differentiating between local and Travis's builds. So locally you can leave tests for longer and occupy use more CPU cores. Travis however requires some gentle treatment - so that's why I was thinking about introducing parameters responsible for time and parallelization etc.

src/libxively/mqtt/codec/xi_mqtt_parser.c
atigyi: yes exactly, or running mqtt parsing on multiple threads

make/mt-config/tests/mt-native
atigyi: ok

src/libxively/mqtt/codec/xi_mqtt_parser.c
olgierdh: or that yes!

src/tests/fuzztests/xi_fuzztest_mqtt_layer.cpp
```diff
@@ -0,0 +1,10 @@
...
+{
+    return 0;
```
olgierdh: ok

src/libxively/mqtt/codec/xi_mqtt_parser.c
DellaBitta: so it's not a bug fix, it's fuzzer support, right?

src/tests/fuzztests/xi_fuzztest_mqtt_layer.cpp
DellaBitta: might want to put a comment here saying "This is the fuzzer signature, and we cannot change it" or some such.

atigyi:
Awesomeness, good to go :+1:

PR #69, opened 2016-12-12
Fix/senml va args argument promotion fix

olgierdh:
[DESCRIPTION]

Using clang-4.0 to compile our library I've noticed one warning about the promotion of types for the va_args arguments is undefined. This is the solution to that problem.


atigyi:
Approved :+1:

milgra:
Approved 👍

PR #71, opened 2016-12-13
[fix] fix for the arm test execution problem

olgierdh:
[DESCRIPTION]

Simple fix for the arm test execution problem after introducing fuzz tests


atigyi:
Approved :ice_skate:

PR #72, opened 2016-12-14
Feature/xcl 2674 travis fix

atigyi:
[ Description ]
False positive builds were experienced on Travis CI. This story investigated the root cause of not triggering red output for failing builds: except command swallowed the failures. Problem solved here, except script now throws problem out to Travis who makes failing builds red again. Hurray.
This "bug" was introduced a while ago (~a month) since then some failures were hidden by these false positives which were unveiled with this move. These bugs were also fixed and now all greens are true-positive builds. Hurray.

[ JIRA ]
find reason of Travis CI's false positive greens
https://jira.3amlabs.net/browse/XCL-2674

[ Reviewer ]

[ QA ]
Eye-o-scopic analysis of each Travis CI build output to check all greens are successful builds. Failing red builds proved that there are true-negative builds as well.

[ Release Notes ]
Travis CI maintenance.


DellaBitta:
Approved

PR #75, opened 2016-12-15
Feature/makefile refactor

atigyi:
[DESCRIPTION]
Some makefile leftovers shreds revisited, removed, rationalised resulting in cleaner less confusing makefiles.

[REVIEWERS]

{JIRA]
[ DEVACCEL ] revisit makefile refactor
https://jira.3amlabs.net/browse/XCL-2664

[QA]
Local dev machine builds work, all Travis CI builds and tests pass.

[RELEASE NOTES]
- makefile maintenance


make/mt-os/mt-linux.mk
```diff
@@ -13,15 +13,12 @@ include make/mt-os/mt-os-common.mk
 
 ifdef XI_SHARED
   AR = gcc
-  XI ?= $(XI_BINDIR)/libxively.so
+  XI = $(XI_BINDIR)/libxively.so
```
DellaBitta: Why remove the optional assignment?  Does it cause a problem?

make/mt-os/mt-osx.mk
```diff
@@ -12,14 +12,13 @@ XI_LIB_FLAGS += $(XI_TLS_LIBFLAGS) -lpthread -lm
 include make/mt-os/mt-os-common.mk
 
 ifdef XI_SHARED
-  XI ?= $(XI_OBJDIR)/libxively.dylib
-  #XI_ARFLAGS := -dynamclib -o $(XI) $(XI_TLS_LIBFLAGS) -L$(dir $(XI_LIBTLS_SHARED))
-  XI_ARFLAGS := -shared -o $(XI) $(XI_TLS_LIBFLAGS) -L$(dir $(XI_LIBTLS_SHARED))
+  XI = $(XI_BINDIR)/libxively.dylib
+  #XI_ARFLAGS := -dynamclib -o $(XI) $(XI_TLS_LIBFLAGS)
```
DellaBitta: remove commented line?

make/mt-os/mt-linux.mk
atigyi: I've made it default to:
XI ?= $(XI_BINDIR)/libxively.a
in mt-os-common.mk. So in case of shared lib we should override.
Having default value avoid the need to set it during a porting process. What do you think is shared lib a practical form, is it needed, who uses this form?

make/mt-os/mt-osx.mk
atigyi: ok, done

PR #77, opened 2017-01-05
Feature/windows commandline compilation stm32

olgierdh:
[DESCRIPTION]

Updates the build system to work against the windows commandline ( git SCM bash ) using STM32Cube SDK and MCU toolchain ( arm-none-eabi-gcc )

[JIRA]
https://jira.3amlabs.net/browse/XCL-2635

[REVIEWERS]

[QA]
All existing tests passes

[RELEASE NOTES]
N/A

### DIFF
--- a/examples/stm32/xively_demo_mt/Src/system_stm32f4xx.c
+++ b/examples/stm32/xively_demo_mt/Src/system_stm32f4xx.c
@@ -1,21 +1,21 @@
 /**
   ******************************************************************************
-  * @file    system_stm32f4xx.c
+  * @file    Demonstrations/Src/system_stm32f4xx.c
   * @author  MCD Application Team
-  * @version V1.3.5
-  * @date    03-June-2016
+  * @version V1.4.4
+  * @date    04-November-2016
   * @brief   CMSIS Cortex-M4 Device Peripheral Access Layer System Source File.
   *
-  *   This file provides two functions and one global variable to be called from 
+  *   This file provides two functions and one global variable to be called from
   *   user application:
-  *      - SystemInit(): This function is called at startup just after reset and 
+  *      - SystemInit(): This function is called at startup just after reset and
   *                      before branch to main program. This call is made inside
   *                      the "startup_stm32f4xx.s" file.
   *
   *      - SystemCoreClock variable: Contains the core clock (HCLK), it can be used
-  *                                  by the user application to setup the SysTick 
+  *                                  by the user application to setup the SysTick
   *                                  timer or configure other parameters.
-  *                                     
+  *
   *      - SystemCoreClockUpdate(): Updates the variable SystemCoreClock and must
   *                                 be called whenever the core clock is changed
   *                                 during program execution.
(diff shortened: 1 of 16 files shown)


make/mt-utils/mt-get-gnu-arm-toolchain.mk
```diff
@@ -22,9 +24,15 @@ else ifeq ($(XI_HOST_PLATFORM),Darwin)
...
 
+	XI_BUILD_PRECONDITIONS := $(CC)
+
+else ifeq ($(XI_HOST_PLATFORM),Windows_NT)
+	CC = arm-none-eabi-gcc
+	AR = arm-none-eabi-ar
 endif
 
-XI_BUILD_PRECONDITIONS := $(CC)
+$(info $$(XI_HOST_PLATFORM)=$(XI_HOST_PLATFORM))
```
olgierdh: Remove that info printout

make/mt-utils/mt-get-gnu-arm-toolchain.mk
```diff
@@ -22,9 +24,15 @@ else ifeq ($(XI_HOST_PLATFORM),Darwin)
 
 	XI_GCC_ARM_TOOLCHAIN_URL := https://launchpad.net/gcc-arm-embedded/5.0/5-2016-q2-update/+download/gcc-arm-none-eabi-5_4-2016q2-20160622-mac.tar.bz2
 
+	XI_BUILD_PRECONDITIONS := $(CC)
```
atigyi: good move man, this means windows is unable to download gnu compiler, right?

examples/stm32/xively_demo_mt/Src/system_stm32f4xx.c
```diff
@@ -1,21 +1,21 @@
 /**
   ******************************************************************************
-  * @file    system_stm32f4xx.c
```
atigyi: was this file update due to an SDK update or similar? so what caused the update of this file and why is it needed, thx.

olgierdh: Yes, this file update is result of updating the SDK - I had compilation error when I used already checked in version of this file. I wonder if we should prepare example per each SDK version ? Or maybe we could use other macro tricks ?

make/mt-utils/mt-get-gnu-arm-toolchain.mk
olgierdh: No, it was my cat :) I hired him to review my code for food.

examples/stm32/xively_demo_mt/Src/system_stm32f4xx.c
DellaBitta: "I wonder if we should prepare example per each SDK version"  

Anyone else nervous about setting a precedent like this? Or is it just me?

olgierdh: Yeah, It's not my best idea... so question here: 

Can we just make it explicit that this example require version x.xx.x of SDK and that's it ?

atigyi: I have a solution offer: we can have links to these files inside the SDK. In case only file contents change and no files are removed, added, then it is a cross-SDK solution which prevents such update needs. So it isn't ultimate solution but one step closer.

atigyi:
Approved 👍

PR #79, opened 2017-01-09
[feat] removed redundant lwipopts.h file

olgierdh:
[DESCRIPTION]

Removes redundant lwipopts.h file which is the lwip tcp/ip stack user configuration for stm32 example project. 

[REVIEWERS]

[JIRA]

https://jira.3amlabs.net/browse/XCL-2642

[QA]

Tested the example application on stm32f41g evalutation board.

[RELEASE NOTES]
N/A


atigyi:
This way each example may configure differently the underlying LwIP library. Isn't a single LwIP setting good for libxively? I mean the default lwipopts.h does not contain the line 
#define LWIP_DNS                        1
which is needed for libxively. This DNS schtuff isn't example specific but libxively specific. I would include the lwipopts.h into our internal sourcebase instead of examples.
What do you think?

olgierdh:
I've updated the location of lwipopts.h. I've moved it to include directory for two reasons:
1) src/bsp didn't look for me as a good candidate as it contains bsp specific implementation and code and I don't want to mix it with the includes/configurations 
2) include folder seems to be a natural environment for this global ( in scope of xively library ) xively specific lwip configuration

atigyi:
I agree that src/bsp is weird and bad location. Include maybe better. What about include/bsp or include/bsp/lwip or include/lwip.

I think we should wait David, maybe he'll have the best idea.

olgierdh:
Ok, I moved the lwipopts.h to the stc/bsp/platform/stm32f4/includes - project compiles and works plus we have only one version of this file. As it is required during the compilation we cannot replace it easily with just defines.

make/mt-os/mt-stm32f4.mk
```diff
@@ -18,6 +18,7 @@ XI_COMPILER_FLAGS += -I$(XI_STM32_PATH_SDK_LWIP)/system
 XI_COMPILER_FLAGS += -I$(XI_STM32_PATH_SDK)/Middlewares/Third_Party/FreeRTOS/Source/include
 XI_COMPILER_FLAGS += -I$(XI_STM32_PATH_SDK)/Middlewares/Third_Party/FreeRTOS/Source/portable/GCC/ARM_CM4F/
 XI_COMPILER_FLAGS += -I$(LIBXIVELY)/examples/stm32/xively_demo_mt/Inc
```
atigyi: off: here libxively build relies on the example's Inc folder which is a unintended dependency. We have to remove it later. But not in this PR since it is different topic, I've just mentioned it for the record.

atigyi:
Approved and thx for iterating on this "minor" issue that much. I think this is the best solution from all. And I think not only lwipopts.h will go into src/bsp/platform/stm32f4/include
Candidates are: FreeRTOSConfig.h, stm32f4xx_hal_conf.h and their friends.

PR #90, opened 2017-01-19
[feat] updated the project settings with the F417

olgierdh:
[DESCRIPTION]
Update of the project settings so it matches the G4 eval board processor type.

[REVIEWERS]

[QA]

[RELEASE NOTES]
N/A


atigyi:
I've also found a file
xively-client-c/examples/stm32/xively_demo_mt/SW4STM32/STM324xG_EVAL/STM32F407IGHx_FLASH.ld 
which isn't 417.

olgierdh:
Yes you are right. Content of this file is exactly the same as the 417 version. That's why I haven't decided to update this file.

atigyi:
Approved 💣 !!

PR #91, opened 2017-01-30
Feature/stm32f4xx wifi sntp

Palantir555:
@milgra All feedback welcome :)


examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
```diff
@@ -0,0 +1,301 @@
...
+   *           - Reference Timestamp 8 bytes
+   *           - Origin Timestamp 8 bytes
+   *           - Receive Timestamp 8 bytes
+   *           - Transmit Timestamp 8 bytes
+   * @param  request_buf is a pointer to a pre-allocated 48 byte char array
+   * @retval void
+   */
+void sntp_build_request( char* request_buf )
+{
+    char sntp_pkt[SNTP_MSG_SIZE] = {
```
atigyi: 2 things here.
1. Why so cumbersome? Why don't we just declare a const char array and pass the address of this to wifi_socket_client_write? Why copying?
2. I don't know the SNTP request UDP message content at all, but looked at file xi_bsp_time_cc3200_sntp.c and there the 48 lenght message is much simpler, pls check it. Although this 2nd is just a note if this works then ok.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
```diff
@@ -0,0 +1,301 @@
...
+   * @retval WiFi_Status_t: WiFi_MODULE_SUCCESS on success, see wifi_interface.h
+   */
+WiFi_Status_t sntp_connect( char* sntp_server, uint32_t sntp_port, uint8_t* sock_id )
+{
+    WiFi_Status_t status = WiFi_MODULE_SUCCESS;
+    uint8_t sntp_protocol = 'u'; //UDP
+    printf("\r\n\tStablishing UDP connection to SNTP server %s:%lu",
+           sntp_server,
+           sntp_port);
+    status = wifi_socket_client_open((uint8_t*)sntp_server, sntp_port,
```
atigyi: Weird this wifi socket API. The UDP is a connection less protocol but we need to call connect on a UDP socket...

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
```diff
@@ -0,0 +1,301 @@
...
+
+/**
+   * @brief
+   * @param  sock_id is the socket ID set by sntp_connect() in stnp_start()
+   * @retval WiFi_Status_t: WiFi_MODULE_SUCCESS on success, see wifi_interface.h
+   */
+WiFi_Status_t sntp_disconnect( uint8_t sock_id )
+{
+    WiFi_Status_t status = WiFi_MODULE_SUCCESS;
+    printf("\r\n>>Closing UDP connection to SNTP server...");
```
atigyi: Yeah, exactly this, the programmer can add only this comment, but there is no such thing: "UDP connection" so cannot close it. But it can stay here because the Wifi API dictates this.

atigyi:
Awesome job JC!

A high level idea should be applied on the location of this solution. So this is an SNTP solution for STM Nucleo + Wifi. Currently resides in the example project. The thing is it should be part of the Xively C Client's STM Nucleo + Wifi solution. Which would mean every further example will benefit from it "freely". Also should be part of Xively C Client because the accurate time is a must for libxively.
So please examine already existing SNTP solution for the CC3200 in directory src/bsp/platform/cc3200. In files: xi_bsp_time_cc3200_sntp.h / .c this solution comes from Bob Burke. And the usage of it is in the TIME BSP: xi_bsp_time_cc3200.c.

Please rename your files to xi_bsp_time_stm32f4_nucleo_wifi_sntp.h / .c (similarly to the aboves) and put them into src/bsp/platform/stm32f4_nucleo_wifi.

Most probably you will have to alter the data received callback and move the implementation of socket differentiation into the BSP's data received impl. Also maybe the new wifi state will be unnecessary.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
Palantir555: 1. You're totally right :)
2. That's the packet recommended by ST in the post they sent us over email. It works fine, so I'll leave it like this

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
Palantir555: I'll change it for 'closing socket to SNTP server'. The WiFi board needs to be notified that we're gonna be reading/writing from a socket, so I think that's where the whole "'connections' in a connection-less protocol" thing comes from. If there was a socket error in the WiFi board, I worried it would be passed down to the Nucleo as a disconnection. I'll fix up the wording in debug messages and whatnot

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Inc/ntp.h
```diff
@@ -0,0 +1,41 @@
...
+    char* response;
+} sntp_response;
+
+/* User API */
+sntp_status_t sntp_get_datetime( uint8_t* socket_id, int32_t* epoch_time );
+void sntp_socket_data_callback( uint8_t socket_id, uint8_t* data_ptr,
+                                uint32_t message_size, uint32_t chunk_size );
+
+/* Internal usage */
+WiFi_Status_t sntp_connect( char* sntp_server, uint32_t sntp_port, uint8_t* sock_id );
```
atigyi: Yes, so these are internal usage functions. Then they shouldn't appear in the header, at least this is my opinion. It just makes the understanding of the behavior harder.
I think here we will need only three functions: 
xi_bspt_time_sntp_init - blocks, makes the UDP request, waits for the answer, parses the answer
xi_bsp_time_sntp_getseconds_posix - returns the current time

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
```diff
@@ -0,0 +1,301 @@
...
+   *         before calling this function
+   * @retval Current date and time in epoch format, -1 on error
+   */
+int32_t sntp_parse_response( char* response )
+{
+    int32_t current_ntp_time = 0; //epoch + 2208988800
+    int32_t current_epoch_time = 0;
+
+    printf("\r\n>>Parsing SNTP response...");
+    memcpy(&current_ntp_time, response+40, 4); //timestamp starts at byte 40
```
olgierdh: Security note: this is a potentially dangerous operation that might be used by someone to do the buffer overflow.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
atigyi: 2. :+1:

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
olgierdh: This +40 makes me scratch my brain :) Could make a const variable that is named like a timestamp offset please :)

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
```diff
@@ -0,0 +1,301 @@
...
+    current_ntp_time = sntp_ntohl(current_ntp_time); //Convert endianness
+    current_epoch_time = current_ntp_time - 2208988800; //Remove NTP offset
```
olgierdh: This big number also - could you please turn it into a const variable ?

atigyi:
Additionally please also handle the BSP TIME module in file xi_bsp_time_stm32f4_nucleo_wifi.c. You can rely on the xi_bsp_time_stm32f4.c file. It is "just" adding up the epoch time at init + uptime.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
```diff
@@ -0,0 +1,301 @@
...
+    }
+    else
+    {
+        printf("\r\n\tReceived epoch time: %ld", sntp_current_time);
+        retval = SNTP_SUCCESS;
+    }
+
+terminate:
+    /* free() SNTP Response */
+    if( NULL != last_sntp_response )
```
olgierdh: Just a minor comment - you could move this logic of checking the NULLness into the implementation of the sntp_free_response - you can also pass the ```**sntp_response``` this way you can set it NULL internally as well. This will increase the readability of this section of the code.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
```diff
@@ -0,0 +1,301 @@
...
+
+/**
+   * @brief
+   * @param  sock_id we're interested in
+   * @retval SNTP_SUCCESS (0) or SNTP_TIMEOUT(-1)
+   */
+sntp_status_t sntp_await_response( uint8_t sock_id )
+{
+    int32_t sntp_timeout_ms = SNTP_TIMEOUT_MS;
+    const int32_t timeout_step = 250;
```
olgierdh: 👍

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
```diff
@@ -0,0 +1,301 @@
...
+{
+    /* Verify we got an SNTP response, not just protocol data */
+    if( (SNTP_MSG_SIZE != message_size) || (SNTP_MSG_SIZE != chunk_size) )
+    {
+        printf("\r\n\tMessage isn't an SNTP respone. It will be ignored");
+        return;
+    }
+
+    /* Store SNTP response */
+    if( NULL != last_sntp_response )
```
olgierdh: Same thing here if you move this check to within the ```sntp_free_response``` this code will become simpler.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
```diff
@@ -0,0 +1,301 @@
...
+    {
+        sntp_free_response(last_sntp_response);
+    }
+    last_sntp_response = calloc(1, sizeof(last_sntp_response));
```
olgierdh: Please use only ```malloc``` and ```memset``` as calloc may not be available on some platforms... yeah I know this is written for this platform but still.... if we just would like to reuse this code somewhere else it would be nice to have such option.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
```diff
@@ -0,0 +1,301 @@
...
+   * @brief  This function shall be called from ind_wifi_socket_data_received(),
+   *         when the sock_id is the one we got from sntp_get_datetime()
+   * @param  See description for ind_wifi_socket_data_received
+   * @retval None
+   */
+void sntp_socket_data_callback(uint8_t sock_id, uint8_t* data_ptr,
+                               uint32_t message_size, uint32_t chunk_size)
+{
+    /* Verify we got an SNTP response, not just protocol data */
+    if( (SNTP_MSG_SIZE != message_size) || (SNTP_MSG_SIZE != chunk_size) )
```
olgierdh: 👍

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
```diff
@@ -0,0 +1,301 @@
...
+/**
+   * @brief
+   * @param  sock_id we're interested in
+   * @retval SNTP_SUCCESS (0) or SNTP_TIMEOUT(-1)
+   */
+sntp_status_t sntp_await_response( uint8_t sock_id )
+{
+    int32_t sntp_timeout_ms = SNTP_TIMEOUT_MS;
+    const int32_t timeout_step = 250;
+    while( NULL == last_sntp_response )
```
olgierdh: There is a probability that this implementation will leak the memory - if the response from the server will come after SNTP_TIMEOUT occurs the ```sntp_socket_data_callback``` will allocate the memory and this memory will not be released - The ```sntp_socket_data_callback``` function must also check if the request is still pending, if not it suppose to send the data to the /dev/null.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
Palantir555: Good point  👍

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Inc/ntp.h
olgierdh: Yes, these should be declared as statics within the *.c file...

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
Palantir555: That buffer overflow is why I wrote the "Length MUST be validated before calling this function" bit. If you look at the caller, we're identifying the message based on its length as reported to the socket data callback. Nothing longer than 48 bytes will reach this function. It felt silly to keep passing the same argument downstream over and over, but is that the recommended approach?

olgierdh: Yes, you are right - the place where you do call this function is protected - that's why I left 👍  there :) This is not something to fix - this is just a note.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/ntp.c
```diff
@@ -0,0 +1,292 @@
...
+   *  - Precision 1 byte
+   *  - Root Delay 4 bytes
+   *  - Root Dispersion 4 bytes
+   *  - Reference Identifier 4 bytes
+   *  - Reference Timestamp 8 bytes
+   *  - Origin Timestamp 8 bytes
+   *  - Receive Timestamp 8 bytes
+   *  - Transmit Timestamp 8 bytes
+   */
+const char SNTP_REQUEST[SNTP_MSG_SIZE] = {
```
olgierdh: If it's internal you can make it static as well.

olgierdh:
Another thing is - I don't think that we want to have these printf's enabled in the release version - so it would be nice to replace these debug log printf's with some macro like we do it for bsp's and xively client library.

Palantir555:
@olgierdh As discussed with Milan, I'll be moving this implementation to the BSP layer after I implement all these changes. I'll use your macros once that's done

olgierdh:
@Palantir555 thanks for the clarification, I can live with that :)

Palantir555:
I've fixed all the issues you mentioned with the code.

I'm currently moving the code into the BSP, and I'll change `printf`s for debug macros immediately after.

Should I merge this PR to make the next one smaller?

Palantir555:
@olgierdh I can't merge the PR myself (lack of write access). If it's good to go, please merge it :)

DellaBitta:
Oh we need to make you a member!

PR #92, opened 2017-01-31
Feature/example firmware update osx

olgierdh:
[DESCRIPTION]
This is very early implementation for firmware updates as an external feature introduced next to the xively-client-library as it's example. This solution is still under the development so the API and it's shape is not final and most probably will be changed.  

[JIRA]
https://jira.3amlabs.net/browse/XCL-2835

[QA]
Basic tests has been performed with the address sanitizer turned on for clang compiler. 

[RELEASE NOTES]
Added example of an implementation for Xively's CPM integrated firmware updates. The reference implementation is not the final one as it's a work in progress and it will change.


examples/firmware_update_example/src/xi_sft.h
```diff
@@ -0,0 +1,131 @@
...
+    XI_FILE_STATUS
+} xi_packet_types;
+
+int xi_sft_init( xi_context_handle_t in_context_handle,
+                 const char* const account_id,
+                 const char* const device_id,
+                 const char* const test_firmware_file );
+void xi_publish_file_info( xi_context_handle_t in_context_handle );
+
+/* functions for platform specific behaviour */
```
olgierdh: One of my conclusions for now is - I'm not sure wether we will be able to make it work in a blocking fashion - like waiting for the library to disconnect may take some time - if we would like to do it gracefully.

examples/firmware_update_example/Makefile
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2003-2016, LogMeIn, Inc. All rights reserved.
```
DellaBitta: it's 2017!

examples/firmware_update_example/src/xi_sft_client.c
```diff
@@ -0,0 +1,674 @@
...
+#include <stdlib.h>
+#include "../cbor/cbor.h"
+#include "../cbor/cn-cbor/cn-cbor.h"
+#include "../sha256/sha256.h"
+#include "xi_sft.h"
+
+/** This file contains the implementation of the firmware update protocol using the cbor
+ * and sha256 library - this is still work in progress and it will be changed */
+
+#ifdef XI_DEBUGSFT
```
DellaBitta: what does DEBUGSFT stand for?

examples/firmware_update_example/src/xi_sft_client.c
```diff
@@ -0,0 +1,674 @@
...
+#define xi_sft_debug_logger( format_string ) __xi_printf( "[SFT] " format_string "\n" )
+#define xi_sft_debug_format( format_string, ... )                                        \
+    __xi_printf( "[SFT] " format_string "\n", __VA_ARGS__ )
+#else /* BSP_DEBUG_LOG */
+#define xi_sft_debug_logger( ... )
+#define xi_sft_debug_format( ... )
+#endif /* BSP_DEBUG_LOG */
+
+
+///// Test these individually
```
DellaBitta: Test them?

examples/firmware_update_example/src/xi_sft_client.c
```diff
@@ -0,0 +1,674 @@
...
+            return;
+    }
+}
+
+void xi_publish_file_info( xi_context_handle_t in_context_handle )
+{
+    unsigned char encoded[512];
+    ssize_t enc_sz;
+
+    // xi_parse_file_update_available(NULL);
```
DellaBitta: should this be removed?

PR #93, opened 2017-01-31
Feature/rationalization of stm32f4xx string usages

atigyi:
[ Description ]
This story try to rationalise the usage of strings: STM32F4, STM32F4xx, STM32 to make usage places informative instead of being confusing. Some places were replaced to STM32Fx saying that the solutions covered are F family independent. At least according to our current knowledge. For instance the LwIP IO NET BSP module is Fx family independent since it is more like network stack specific (LwIP). The random solution is also more like hal_rng dependent than Fx. 

[ JIRA ]
STM32Fx consider renaming strings in make files to STM32Fx
https://jira.3amlabs.net/browse/XCL-2649

[ Reviewer ]

[ QA ]
Both STM32F4 eval and Nucleo projects were built. Would be good to cover F7 and other F boards as well but we are not there yet.

[ Release Notes ]
n/a


.travis.yml
```diff
@@ -12,10 +12,9 @@ env:
     - PRESET=ARM_REL
     - PRESET=ARM_REL_MIN
 
-    - PRESET=STM32F4
+    - PRESET=STM32FX
```
olgierdh: This is better - I wonder if it's true and if this example will work for entire FX family ?

examples/stm32/xively_demo_mt/SW4STM32/STM324xG_EVAL/.cproject
```diff
@@ -39,8 +39,8 @@
 								<option id="gnu.c.compiler.option.include.paths.389311824" name="Include paths (-I)" superClass="gnu.c.compiler.option.include.paths" useByScannerDiscovery="false" valueType="includePath">
 									<listOptionValue builtIn="false" value="../../../Inc"/>
 									<listOptionValue builtIn="false" value="${XI_CLIENT_PATH}/include"/>
-									<listOptionValue builtIn="false" value="&quot;${XI_CLIENT_PATH}/include/bsp&quot;"/>
-									<listOptionValue builtIn="false" value="&quot;${XI_CLIENT_PATH}/src/bsp/platform/stm32f4/includes&quot;"/>
+									<listOptionValue builtIn="false" value="${XI_CLIENT_PATH}/include/bsp"/>
+									<listOptionValue builtIn="false" value="${XI_CLIENT_PATH}/src/bsp/platform/stm32fx/include"/>
```
olgierdh: 👍

make/mt-config/mt-presets.mk
```diff
@@ -99,19 +99,19 @@ else ifeq ($(PRESET), CC3200_TLS_SOCKET)
     XI_TARGET_PLATFORM = cc3200
 
 # -------------------------------------------------------
-# ST Micro STM32F4
-else ifeq ($(PRESET), STM32F4)
-    CONFIG = $(CONFIG_STM32F4)
+# ST Micro STM32FX
```
DellaBitta: I'm a bit confused. There's a Generic STM32FX build configuration, and a more specific STM32FX Nucleo Wifi configuration which is very specific.

When would the STM32FX build configuration be applicable?  Or is this just a generic work in progress for our next STM32F platform?

atigyi: good point. It can be good for Eval and Discovery. We don't have info about specifics now. This is the status of our knowledge and development process. But yes I agree, it is a bit odd, but we don't know more currently.

.travis.yml
atigyi: I believe this is more generic than specific. E.g. the LwIP solution is not F family specific. Neither the memory, rng.

PR #94, opened 2017-01-31
[WIP] initial commit

atigyi:


make/mt-os/mt-stm32f4_nucleo_wifi.mk
```diff
@@ -16,6 +16,8 @@ XI_COMPILER_FLAGS += -I$(XI_STM32_PATH_SDK)/Drivers/CMSIS/Device/ST/STM32F4xx/In
 XI_COMPILER_FLAGS += -I$(XI_STM32_PATH_SDK)/Drivers/STM32F4xx_HAL_Driver/Inc
 XI_COMPILER_FLAGS += -I$(XI_STM32_PATH_SDK)/Drivers/BSP/STM32F4xx-Nucleo
 
+XI_COMPILER_FLAGS += -I$(LIBXIVELY)/examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Inc
```
atigyi: last spring Kotyo worked on cut the dependency between Xively C Client and its examples. Reason is that a nucleo wifi libxively build can have multiple example using it. Thus the libxively cannot depend on their usage places.

atigyi: I think you should add this include path instead STM32CubeExpansion_WIFI1_V2.1.1/Projects/Multi/Applications/Client_Socket/Inc
as I remember Kotyo solved this issue with adding the following to the include path
XI_COMPILER_FLAGS += -I$(XI_STM32_PATH_SDK)/Projects/STM324xG_EVAL/Templates/Inc

Because this - I guess - is all about having the the board specific HAL config file in the include path: stm32f4xx_hal_conf

PR #97, opened 2017-02-06
[fix] fix for nucleo bsp implementation

olgierdh:
[DESCRIPTION]

Fixes the problem with receiving both callbacks from the WiFi Shield and a problem with not clearing the local memory pointers.

[QA]

Manual tests where connecting with correct and incorrect credentials. 

[RELEASE NOTES]

Fixed bsp network implementation for ST NUCLEO F4x boards issues:
i) receiving data and connection shutdown issue
ii) re-connecting to the broker

### DIFF
--- a/src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
+++ b/src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
@@ -7,6 +7,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <xi_bsp_io_net.h>
+#include <assert.h>
 #include "xi_bsp_debug.h"
 #include <xi_data_desc.h>
 #include "wifi_interface.h"
@@ -14,7 +15,7 @@
 
 #ifdef __cplusplus
 extern "C" {
-#endif 
+#endif
 #ifndef MAX
 #define MAX( a, b ) ( ( ( a ) > ( b ) ) ? ( a ) : ( b ) )
 #endif
@@ -30,14 +31,20 @@ typedef struct _xi_bsp_stm_net_state_t
 {
     xi_data_desc_t* head;
     xi_bsp_stm_wifi_state_t wifi_state;
-
 } xi_bsp_stm_net_state_t;
 
-xi_bsp_stm_net_state_t xi_net_state = {0};
+/* this is local variable */
+static xi_bsp_stm_net_state_t xi_net_state = {NULL, 0};
 
 xi_bsp_io_net_state_t xi_bsp_io_net_create_socket( xi_bsp_socket_t* xi_socket )
 {
+    /* sanity check */
+    assert( NULL == xi_net_state.head );
+
     *xi_socket = 0;
+
+    /* reset the wifi state */
+    xi_net_state.wifi_state = xi_wifi_state_connected;
 
     return XI_BSP_IO_NET_STATE_OK;
 }
@@ -119,6 +126,10 @@ xi_bsp_io_net_state_t xi_bsp_io_net_read( xi_bsp_socket_t xi_socket,
             xi_data_desc_t* head = xi_net_state.head;
 
             const size_t bytes_available = head->length - head->curr_pos;
+
+            /* sanity check */
+            assert( 0 != bytes_available );
+
             const size_t bytes_to_copy =
                 ( bytes_available > count ) ? count : bytes_available;
 
@@ -138,7 +149,7 @@ xi_bsp_io_net_state_t xi_bsp_io_net_read( xi_bsp_socket_t xi_socket,
     {
         return XI_BSP_IO_NET_STATE_ERROR;
     }
-    else if ( xi_net_state.wifi_state == xi_wifi_state_connected )
+    else if ( xi_net_state.wifi_state == xi_wifi_state_disconnected )
     {
         return XI_BSP_IO_NET_STATE_CONNECTION_RESET;
     }
@@ -152,14 +163,17 @@ xi_bsp_io_net_state_t xi_bsp_io_net_close_socket( xi_bsp_socket_t* xi_socket )
     status               = wifi_socket_client_close( *xi_socket );
 
     /* free all buffers */
-
     xi_data_desc_t* head = xi_net_state.head;
+
     while ( NULL != head )
     {
         xi_data_desc_t* temp = head;
         head                 = head->__next;
         xi_free_desc( &temp );
     }
+
+    /* reset it to the default value after deleting all elements */
+    xi_net_state.head = NULL;
 
     if ( status == WiFi_MODULE_SUCCESS )
     {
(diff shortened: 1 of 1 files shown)


src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
```diff
@@ -30,15 +31,21 @@ typedef struct _xi_bsp_stm_net_state_t
...
 
 xi_bsp_io_net_state_t xi_bsp_io_net_create_socket( xi_bsp_socket_t* xi_socket )
 {
+    /* sanity check */
+    assert( NULL == xi_net_state.head );
+
     *xi_socket = 0;
 
+    /* reset the wifi state */
+    xi_net_state.wifi_state = xi_wifi_state_connected;
```
milgra: 👍 nice one! I forgot to re-initialize in case of a reconnection

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
```diff
@@ -138,7 +149,7 @@ xi_bsp_io_net_state_t xi_bsp_io_net_read( xi_bsp_socket_t xi_socket,
     {
         return XI_BSP_IO_NET_STATE_ERROR;
     }
-    else if ( xi_net_state.wifi_state == xi_wifi_state_connected )
+    else if ( xi_net_state.wifi_state == xi_wifi_state_disconnected )
```
milgra: 👍 I totally missed that :(

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
```diff
@@ -152,15 +163,18 @@ xi_bsp_io_net_state_t xi_bsp_io_net_close_socket( xi_bsp_socket_t* xi_socket )
...
+
     while ( NULL != head )
     {
         xi_data_desc_t* temp = head;
         head                 = head->__next;
         xi_free_desc( &temp );
     }
 
+    /* reset it to the default value after deleting all elements */
+    xi_net_state.head = NULL;
```
milgra: 👍

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
atigyi: we consider the socket "connected" right after the BSP create?

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
atigyi: 👍

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
atigyi: 👍 maybe this could be a list function / macro in the xi_list.h

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
```diff
@@ -179,37 +193,30 @@ xi_bsp_io_net_state_t xi_bsp_io_net_select( xi_bsp_socket_events_t* socket_event
...
-    else
-    {
-        return XI_BSP_IO_NET_STATE_ERROR;
+        if ( 1 == socket_events->in_socket_want_write )
+        {
+            socket_events->out_socket_can_write = 1;
+        }
     }
+
+    return XI_BSP_IO_NET_STATE_OK;
```
atigyi: this select does not do passive waiting. Is this done in the application?

atigyi: just asking and no change request for this here and now: 
So in theory we could remove the HAL_Delay from the wifi_state_idle: state and put this delay conditionally here into the select function. And call this delay only if there is nothing to do: read, write, connect.
Is this true?

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
olgierdh: @atigyi - if you take a look at the ```xi_net_state.wifi_state``` value at line 37 - the value 0 is xi_wifi_state_connected - so if we reach this function - we assumed already that it's state was connected - It's true that it is misleading - I'll add another state for uninitialised value - this way our code will be more bulletproof - 💥

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
olgierdh: No worries I missed that as well during the rewiew - I guess having unit tests or integration tests or functional tests would help us a lot during porting - it would speed up developing a production quality code.

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
olgierdh: Yeah there is a xi_list.h implementation - I think Milan was trying not add another dependency - I'll convert it to use it. That's a good idea @atigyi.

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
olgierdh: Yes that's true. I was thinking about it - but at the same time we don't want to block the user application for longer that it is necessary - this not having select situation is really problematic and we have to carefully think about it - because of few facts:
Fact no 1. - if we wait for too long - the WiFi cheap may read too much data and it may say - no more memory closing 
Fact no 2. - if we wait for too long - the WiFi cheap may read big amount of data - and before Xively will consume it it will fill our heap with chunks 
Fact no 2. - if we wait for too short - we will consume more power.... in MQTT protocol there is no way to save power or go to sleep mode between sending and receiving messages but the MQTT-SN is ready for it. 

So with that being said I think we should aim into something between ( 50ms - 250ms ) question what will work the best if users want to implement their own activities - I think that it will be much more clear once we add the sensor board and CPM connection so we will be able to see what is the response time of the demo application.

atigyi: ok, thx the answer

atigyi:
Approved!

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
```diff
@@ -17,15 +17,18 @@
...
 #endif
+
 #ifndef MAX
 #define MAX( a, b ) ( ( ( a ) > ( b ) ) ? ( a ) : ( b ) )
 #endif
 
+#ifndef MIN
+#define MIN( a, b ) ( ( ( a ) < ( b ) ) ? ( a ) : ( b ) )
+#endif
+
```
milgra: 👍

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
```diff
@@ -35,7 +38,7 @@ typedef struct _xi_bsp_stm_net_state_t
 } xi_bsp_stm_net_state_t;
 
 /* this is local variable */
-static xi_bsp_stm_net_state_t xi_net_state = {NULL, 0};
+static xi_bsp_stm_net_state_t xi_net_state = {NULL, xi_wifi_state_disconnected};
```
milgra: 👍

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
```diff
@@ -7,36 +7,44 @@
...
 typedef struct _xi_bsp_stm_net_state_t
 {
     xi_data_desc_t* head;
     xi_bsp_stm_wifi_state_t wifi_state;
-
 } xi_bsp_stm_net_state_t;
 
-xi_bsp_stm_net_state_t xi_net_state = {0};
+/* this is local variable */
+static xi_bsp_stm_net_state_t xi_net_state = {NULL, xi_wifi_state_disconnected};
```
atigyi: 👍

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
```diff
@@ -53,6 +61,8 @@ xi_bsp_io_net_connect( xi_bsp_socket_t* xi_socket, const char* host, uint16_t po
 
     if ( status == WiFi_MODULE_SUCCESS )
     {
+        /* reset the wifi state */
+        xi_net_state.wifi_state = xi_wifi_state_connected;
```
atigyi: 👍

atigyi:
Approved #2.

PR #98, opened 2017-02-08
Feature/port eth nucleo f429 zi

atigyi:
[ Description ]
A new Eclipse System Workbench project is added to examples/stm32 directory: xively_demo_stm32f4xx_nucleo_eth. This project targets the STM32F429ZI Ethernet board for the F4 Nucleo. After building the Xively C Client with PRESET=STM32FX and updating 3 environment variables in the Eclipse project the build emits a binary targeting the F429ZI board. A good to know fact is that the Xively C Client's format is almost identical with the F417GI Eval board's format. "Only" a preprocessor definition is changed which is the MCU type.
The F4 Eth Eclipse project contains the MEMS sensor board logic. The code is duplicated from the wifi example.
This change also contains an initial PoC example SW project for F2 ethernet board.

[ JIRA ]
STM32F4xx Nucleo wired Ethernet connected to Xively
https://jira.3amlabs.net/browse/XCL-2827

[ Reviewer ]

[ QA ]
Xively C Client lib was built. Eclipse project for WolfSSL and the new project was built, output deployed on the board. The connect callback returned with success.

[ Release Notes ]
N/A


examples/stm32/xively_demo_stm32f4xx_nucleo_eth/Src/xively_client.c
```diff
@@ -0,0 +1,906 @@
...
+/*
+ *  Device Specific
+ *  Credential Information
+ */
+#define MS_DEVICE 0
+#define BB_DEVICE 1
+
+#define BUILD_DEVICE BB_DEVICE
+
+#define XC_ACCOUNT_ID "bc4ebe58-b070-437b-9a64-a112bb2283d8"
```
olgierdh: I don't think that we should put these credentials into the public repository... but haven't we putted them already ?

make/mt-os/mt-stm32fx.mk
```diff
@@ -44,7 +44,8 @@ XI_COMPILER_FLAGS += -ffunction-sections
 # loop optimization
 XI_COMPILER_FLAGS += -fno-move-loop-invariants
 # define the board type
-XI_COMPILER_FLAGS += -DSTM32F407xx
+# XI_COMPILER_FLAGS += -DSTM32F407xx
+XI_COMPILER_FLAGS += -DSTM32F429xx
```
olgierdh: So this is the guy we have to change every time we need different board.... What to do with that guy ? He is not very handy.... Maybe we could pass this as a makefile variable ? Like XI_BOARD_SYMBOL ?= STM32F407xx and then we could overwrite it ? Dunnno. But it seems that we have to be able to handle this somehow. Another idea is to use separate board-focused.mk files and include them from here using name of the board or smth. I know it rises the complexity but we need a way to handle different board versions.

examples/stm32/xively_demo_stm32f4xx_nucleo_eth/Src/xively_client.c
atigyi: ok, removed, thx

Palantir555:
The PR looks great :+1:

I'm still a bit worried about making the code unnecessarily complex. Bob's client interface is fantastic, but not exactly straightforward for new users. That being said, we already have an ST example with reasonably simple code, so I have no problem keeping these ones as examples of more versatile implementations. A thought: We might wanna let the devs know, though (comment/example-specific Readme.md/??) so they know to check out other examples if they get discouraged.

There are some "morningstar" credentials hardcoded in there -- we gotta figure out where they come from and get rid of them. Other than that, we can take care of all the comments and code cleanup later. I'm OK with merging this so we can start working on top of it.

DellaBitta:
I'm going to spend some time cleaning up xively_main.c over the next day ot make it fall more in line with the new wifi main.c, hopefully simplifying it.  

We will definite need to iterate on the source over the next few weeks, but that shouldn't hold up this PR and the tutorial from being written.

examples/stm32/xively_demo_stm32f2xx_nucleo_eth/Inc/stm32f2xx_hal_conf.h
```diff
@@ -0,0 +1,402 @@
...
+#define HAL_SPI_MODULE_ENABLED   
+#define HAL_TIM_MODULE_ENABLED   
+#define HAL_UART_MODULE_ENABLED 
+#define HAL_USART_MODULE_ENABLED 
+#define HAL_IRDA_MODULE_ENABLED 
+#define HAL_SMARTCARD_MODULE_ENABLED 
+#define HAL_WWDG_MODULE_ENABLED  
+#define HAL_CORTEX_MODULE_ENABLED
+#define HAL_PCD_MODULE_ENABLED
+#define HAL_HCD_MODULE_ENABLED
```
atigyi: Checked it, removed those which are removed in the wifi example. The binaries were same sized compared to the more enabled module one. So no benefit. But the button wasn't working. So I wouldn't bother these if it is ok.

examples/stm32/xively_demo_stm32f2xx_nucleo_eth/SW4STM32/STM32F207ZG_NUCLEO_144/.project
```diff
@@ -0,0 +1,495 @@
...
+			<locationURI>PARENT-2-PROJECT_LOC/Src/stm32f2xx_it.c</locationURI>
+		</link>
+		<link>
+			<name>Application/User/xively_client.c</name>
+			<type>1</type>
+			<locationURI>PARENT-2-PROJECT_LOC/Src/xively_client.c</locationURI>
+		</link>
+		<link>
+			<name>Drivers/BSP/Components</name>
+			<type>2</type>
```
atigyi: Very good catch, thx, fixing!!

atigyi: done

examples/stm32/xively_demo_stm32f2xx_nucleo_eth/Src/xively_client.c
```diff
@@ -0,0 +1,1214 @@
...
+                if (report_data[e] != ms_data[MS_NOW_V][e])
+                  {
+                    report_data[e] = ms_data[MS_NOW_V][e];
+                    report_pending[e] = 1;
+                  }
+              }
+          }
+      }
+  }
+#endif
```
atigyi: removed in both xively_client.c files

examples/stm32/xively_demo_stm32f2xx_nucleo_eth/Src/xively_client.c
```diff
@@ -0,0 +1,1214 @@
...
+static const mqtt_topic_configuration_t topics_array[] = {
+    {{XI_TOPIC_NAME_MANGLE( "Accelerometer" )}, NULL, pub_accelerometer},
+    {{XI_TOPIC_NAME_MANGLE( "Magnetometer" )}, NULL, pub_magnetometer},
+    {{XI_TOPIC_NAME_MANGLE( "Barometer" )}, NULL, pub_barometer},
+    {{XI_TOPIC_NAME_MANGLE( "Gyroscope" )}, NULL, pub_gyroscope},
+    {{XI_TOPIC_NAME_MANGLE( "Humidity" )}, NULL, pub_humidity},
+    {{XI_TOPIC_NAME_MANGLE( "Button" )}, NULL, pub_button},
+    {{XI_TOPIC_NAME_MANGLE( "LED" )}, &on_led_msg, NULL},
+    {{XI_TOPIC_NAME_MANGLE( "Temperature" )}, NULL, pub_temperature},
+};
```
atigyi: yeah, I think we should refactor this and use common solution for both projects: eth and wifi. But for now let's keep it duplicated since no resources to do the job. Maybe after the Embedded World Demo?

examples/stm32/xively_demo_stm32f4xx_nucleo_eth/Src/main.c
```diff
@@ -0,0 +1,367 @@
...
+         - Configure the Systick to generate an interrupt each 1 msec
+         - Set NVIC Group Priority to 4
+         - Global MSP (MCU Support Package) initialization
+       */
+    HAL_Init();
+
+    /* Configure the system clock to 180 MHz */
+    SystemClock_Config();
+
+    BSP_LED_Init( LED3 );
```
atigyi: Sorry but how things seem now this won't be priority. But I would prefer the first because it will be enough for the next 1 year. But if wind turns we can implement generic Xively Demo GPIO handling mechanism.

examples/stm32/xively_demo_stm32f4xx_nucleo_eth/Src/xively_client.c
```diff
@@ -0,0 +1,1219 @@
...
+ *  Used by Bob Burke, Xively Professional Services
+ */
+#elif BUILD_DEVICE == BB_DEVICE
+#define XI_ACCOUNT_ID ""
+#define XI_DEVICE_ID ""
+#define XI_PASSWORD ""
+
+#else
+#error Invalid "BUILD_DEVICE"
+#endif
```
atigyi: removing them

examples/stm32/xively_demo_stm32f2xx_nucleo_eth/Inc/stm32f2xx_hal_conf.h
Palantir555: I'm curious why they use these defines in the SDK, then :thinking: Hopefully they're not powering up or initialising any peripherals simply because of these defines.

If using the wifi ones caused compilation to fail, there must be different peripherals for the 2 boards, or a missing one or something. We can debug this when we have more time; I have no problem leaving them enabled for now (until we find out what exactly they're supposed to do).

PR #101, opened 2017-02-10
[fix] fixes for sensor demo compilation

olgierdh:
[DESCRIPTION]

Fixed some developer specific settings and made them to be path dependent.


Palantir555:
Looks great, thanks for the fix :ok_hand:

*Mental note: Don't trust Eclipse's drag'n'drop anymore, even though it asks which path you wanna make new files relative to. `git grep palantir` before every PR :)*

PR #104, opened 2017-03-02
Functional TLS validation and rotating SNTP servers

Palantir555:
[DESCRIPTION]
TLS certificates are now validated. The WiFi module only supports storage of 1 root CA certificate at a time, but I've included all of them in the header file in case they're useful in the future or for our users.

Implement rotating SNTP servers-- Whenever a device sends an SNTP request, the address they're supposed to connect to is set to the next one in the list. That way, no server is hit twice consecutively by the same device unless it's rebooted. This stops the boot process from being halted due to SNTP server downtime.

[REVIEWERS]

[JIRA]
Implement TLS Certificate Validation (Task: ???)

[QA]
For TLS Validation:
- Successfully connected to xively broker using Globalsign root cert
- Failed to connect to xively broker using a different root cert
- Failed to connect to broker.xively.com passing an incorrect server address for validation
- Failed to connect to a 3rd party broker using the Globalsign root cert (they use a different CA)

For SNTP server rotation:
- Reboot the device when the first SNTP server is down, review the logs to see it fail, try the next server and succeed
- Hardcode a "return ERROR;" in the `xi_bsp_time_sntp_init()` function, causing the loop in `xi_bsp_time_init()` to continuously iterate without delay, displaying all server addresses in correct order in the debug logs.

[RELEASE NOTES]
N/A


Palantir555:
I ran these tests to make sure TLS validation was working as intended:

1. Successfully connect to xively broker using Globalsign root cert
2. Fail to connect to xively broker using a different root cert
3. Fail to connect to broker.xively.com passing an incorrect server address for validation
4. Fail to connect to a 3rd party broker using the Globalsign root cert (they use a different CA)

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_tls_certs.h
```diff
@@ -0,0 +1,101 @@
+#define GLOBALSIGN_ROOT_CERT "-----BEGIN CERTIFICATE-----\n\
```
DellaBitta: Are these certificates part of the xively library, too?  We might be doubling up the storage space here.

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
```diff
@@ -50,12 +51,39 @@ xi_bsp_io_net_state_t xi_bsp_io_net_create_socket( xi_bsp_socket_t* xi_socket )
...
+    uint8_t* tls_mode             = ( uint8_t* )"o"; //["m"utual || "o"ne-way]
+    char* tls_cert                = GLOBALSIGN_ROOT_CERT;
+
+    xi_bsp_debug_format( "Trusted Root CA Certificate:\r\n%s", tls_cert );
+    status =
+        wifi_socket_client_security( tls_mode, ( uint8_t* )tls_cert, NULL, NULL,
+                                     ( uint8_t* )host, ( uint32_t )current_datetime );
+
+    if ( status != WiFi_MODULE_SUCCESS )
+    {
```
DellaBitta: we had a standard of doing ( Wifi_Fi_MODULE_SUCCESS != status ) just to be safe and not accidentally assign to the status variable instead of comparing against it.

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
```diff
@@ -50,12 +51,39 @@ xi_bsp_io_net_state_t xi_bsp_io_net_create_socket( xi_bsp_socket_t* xi_socket )
...
 }
 
+static xi_bsp_io_net_state_t xi_bsp_io_net_configure_tls( const char* host )
+{
+    WiFi_Status_t status          = WiFi_MODULE_SUCCESS;
+    posix_time_t current_datetime = xi_bsp_time_sntp_getseconds_posix();
+    uint8_t* tls_mode             = ( uint8_t* )"o"; //["m"utual || "o"ne-way]
+    char* tls_cert                = GLOBALSIGN_ROOT_CERT;
+
+    xi_bsp_debug_format( "Trusted Root CA Certificate:\r\n%s", tls_cert );
```
DellaBitta: Does this print the entire certificate to the console?  Might want to drop this from the actual check?

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
```diff
@@ -50,12 +51,39 @@ xi_bsp_io_net_state_t xi_bsp_io_net_create_socket( xi_bsp_socket_t* xi_socket )
     return XI_BSP_IO_NET_STATE_OK;
 }
 
+static xi_bsp_io_net_state_t xi_bsp_io_net_configure_tls( const char* host )
+{
+    WiFi_Status_t status          = WiFi_MODULE_SUCCESS;
+    posix_time_t current_datetime = xi_bsp_time_sntp_getseconds_posix();
+    uint8_t* tls_mode             = ( uint8_t* )"o"; //["m"utual || "o"ne-way]
+    char* tls_cert                = GLOBALSIGN_ROOT_CERT;
+
```
DellaBitta: does this mean that we have to add logic to attempt connections with different certificates if the first attempt fails?  I think Bob did something similar for a device at one point but I don't know if we've ever really thought through what this really should look like.

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_time_stm32f4_nucleo_wifi_sntp.c
```diff
@@ -121,29 +126,35 @@ static void sntp_free_response( sntp_response_t** r )
...
-   *               rotate through them, changing to the next one every time this
-   *               function is called
    */
 static WiFi_Status_t sntp_start( uint32_t sntp_port, uint8_t* sock_id )
 {
-    WiFi_Status_t status = WiFi_MODULE_SUCCESS;
-    uint8_t sntp_protocol = 'u'; //UDP
-    if( NULL == sock_id )
+    WiFi_Status_t status                  = WiFi_MODULE_SUCCESS;
+    uint8_t sntp_protocol                 = 'u'; // UDP
```
DellaBitta: standard is to use /* */ instead of //

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_time_stm32f4_nucleo_wifi_sntp.c
```diff
@@ -6,6 +6,11 @@
 #include "wifi_interface.h"
 #include "xi_bsp_time_stm32f4_nucleo_wifi_sntp.h"
 
+#define SNTP_PORT   123
+#define SNTP_SERVER_HOST_LEN 15
+char* sntp_servers[SNTP_SERVER_HOST_LEN] = {"0.pool.ntp.org", "1.pool.ntp.org",
```
atigyi: array size SNTP_SERVER_HOST_LEN ?
not 4?

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_io_net_stm32fx_nucleo_wifi.c
```diff
@@ -50,12 +51,39 @@ xi_bsp_io_net_state_t xi_bsp_io_net_create_socket( xi_bsp_socket_t* xi_socket )
     return XI_BSP_IO_NET_STATE_OK;
 }
 
+static xi_bsp_io_net_state_t xi_bsp_io_net_configure_tls( const char* host )
+{
+    WiFi_Status_t status          = WiFi_MODULE_SUCCESS;
+    posix_time_t current_datetime = xi_bsp_time_sntp_getseconds_posix();
+    uint8_t* tls_mode             = ( uint8_t* )"o"; /* ["m"utual || "o"ne-way] */
```
atigyi: from where have you figured out this ["m"utual || "o"ne-way] ?

atigyi: syntax: pls add const where possible to give hint to reader not to expect value change, thx

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_tls_certs.h
atigyi: yeah, probably JC took this from xi_RootCA_list.c, right JC?

1. in file xi_RootCA_list.c these are concatenated together into a single array since wolfSSL is able to handle multiple Root CAs. Here we could try whether the wifi module handles multiple TLS certs passed to API, could you test this JC please?

2. Does not double the storage for two reasons: a) the xi_RootCA_list array is not compiled in here, because no TLS layer is compiled in either because the TLS is done by the wifi chip. b) is my guess that even if it it was included the compiler would drop it out if it isn't used, but this is up to the compiler.

3. Do you know why we initialize the xi_RootCA_list with hexa numbers? I mean we could initialize the array with human-readable string, with the same string JC uses here: PEM format cert.

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_time_stm32f4_nucleo_wifi_sntp.c
```diff
@@ -6,6 +6,11 @@
 #include "wifi_interface.h"
 #include "xi_bsp_time_stm32f4_nucleo_wifi_sntp.h"
 
+#define SNTP_PORT   123
+#define SNTP_SERVER_HOST_LEN 15
+char* sntp_servers[] = {"0.pool.ntp.org", "1.pool.ntp.org",
```
atigyi: in xi_bsp_time_cc3200_sntp.c Bob uses these servers
"pool.ntp.org", "time-a.nist.gov", "time-b.nist.gov", "time-c.nist.gov"
These four sound more ... hm ... reliable, or might have higher aggregated uptime compared to *.pool.ntp.org.

Just guessing, so if you don't think I am ok with your selection.

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_time_stm32f4_nucleo_wifi_sntp.c
```diff
@@ -121,29 +126,35 @@ static void sntp_free_response( sntp_response_t** r )
...
-   *               function is called
    */
 static WiFi_Status_t sntp_start( uint32_t sntp_port, uint8_t* sock_id )
 {
-    WiFi_Status_t status = WiFi_MODULE_SUCCESS;
-    uint8_t sntp_protocol = 'u'; //UDP
-    if( NULL == sock_id )
+    WiFi_Status_t status                  = WiFi_MODULE_SUCCESS;
+    uint8_t sntp_protocol                 = 'u'; /* UDP */
+    static const uint8_t sntp_servers_num = sizeof( sntp_servers ) / SNTP_SERVER_HOST_LEN;
```
atigyi: something stinks here. I think
sizeof( sntp_servers ) equals to 4 x sizeof( char* ) which is 16 on a 32 byte system. 16 / 15 is 1. So this will check "only" the first two servers. At least this is my bet.

I think you could use Bob's solution from the file xi_bsp_time_cc3200_sntp.c
#define NTP_SERVER_COUNT ( sizeof( ntp_server_names ) / sizeof( char* ) )

atigyi: Btw the problem originates from copying the sntp implementation and haven't refactored it and using common logic from same file. I did this.

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_time_stm32f4_nucleo_wifi_sntp.c
```diff
@@ -121,29 +126,35 @@ static void sntp_free_response( sntp_response_t** r )
...
+    static uint8_t sntp_server_turn       = 0;
+    char* sntp_server                     = sntp_servers[sntp_server_turn];
+
+    /* Use a different SNTP server every time this function is called */
+    sntp_server_turn++;
```
atigyi: why don't we step to the next server only if the current one fails?

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_time_stm32f4_nucleo_wifi_sntp.c
```diff
@@ -121,29 +126,35 @@ static void sntp_free_response( sntp_response_t** r )
...
-   * @TODO: sntp.h should have a list of SNTP servers. This function should
-   *               rotate through them, changing to the next one every time this
-   *               function is called
    */
 static WiFi_Status_t sntp_start( uint32_t sntp_port, uint8_t* sock_id )
 {
-    WiFi_Status_t status = WiFi_MODULE_SUCCESS;
-    uint8_t sntp_protocol = 'u'; //UDP
-    if( NULL == sock_id )
+    WiFi_Status_t status                  = WiFi_MODULE_SUCCESS;
```
atigyi: Also here this nucleo wifi implementation calls this sntp_start function once. Because the xi_bsp_time_sntp_init is called once. So it tries only the first server. the sntp_start's implementation expects the caller to handle failed connection but in this case the caller does not handle error.

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_tls_certs.h
atigyi: oh, I answered this question already, maybe on the main page of the PR, but apparently it is not here.

For now I am satisfied with these certs here. I've asked question regarding using multiple certs in a single array, we'll get our answers or if you dare JC you could experiment it.

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_time_stm32f4_nucleo_wifi_sntp.c
atigyi: I just realized that this is not true. As I see the xi_bsp_time_init for the wifi calls this time acquisition until it succeeds. So this is ok for now.

PR #105, opened 2017-03-02
Stm32f4 nucelo publish rate increase

DellaBitta:
[ Description ]
Updates to the STM32F4 Nucelo demo application in preparation for Embedded World 2017 showing.

The following changes were made to main.c for the STM32F4 Wifi Example:

* Removed Topic Table structure.  This added reading complexity and while adding flexibility, the demo is rather static.
* Added functionality to track connection state.
* Cleaned up some serial output to make it more readable. 
* Created a new BURST mode where Accelerometer, Gyroscope and Button status topics are published to in high frequency when pushing the blue button on the device. This mode can be switched off by pushing the blue button again.
* Ran the Xively code formatter to strip tabs and make things consistent.
* Added Xively copyright.

[ JIRA ]
None.

[ Reviewer ]
@atigyi 
@Palantir555 

[ QA ]
* Tested application over the course of a 24 hour window to ensure that it was operational throughout.
* Toggled Burst Mode on and off and watched the results pour through the Xively Messaging Widget on CPM.
* Ensured that all topics were publishing correctly by checking Messaging Widget.
* Ensured that LED control was still working by publishing to the subscribed LED topic.
* Tested with embedded Wifi Credentials and with WiFi credentials entered into UART.

[ Release Notes ]
None  (STM32F4 demo isn't yet released)

### DIFF
--- a/examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
+++ b/examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
@@ -82,6 +82,7 @@ WiFi_Status_t wifi_get_AP_settings( void );
 wifi_state_t wifi_state;
 wifi_config config;
 wifi_scan net_scan[WIFI_SCAN_BUFFER_LIST];
+xi_state_t xi_connection_state = XI_SOCKET_NO_ACTIVE_CONNECTION_ERROR;
 
 char console_ssid[40];
 char console_psk[20];
@@ -90,6 +91,7 @@ char* ssid          = "SSID";
 char* seckey        = "PASSWORD";
 WiFi_Priv_Mode mode = WPA_Personal;
 
+
 #define XI_ACCOUNT_ID "XIVELY ACCOUNT ID"
 #define XI_DEVICE_ID "XIVELY DEVICE ID"
 #define XI_DEVICE_PASS "XIVELY DEVICE PASSWORD"
@@ -97,22 +99,17 @@ WiFi_Priv_Mode mode = WPA_Personal;
 /* the interval for the time function */
 #define XI_PUBLISH_INTERVAL_SEC 2
 
-typedef struct topic_descr_s
-{
-    const char* const name;
-} mqtt_topic_descr_t;
-
 /* declaration of fn pointer for the msg handler */
 typedef void ( *on_msg_handler_fn_t )( const uint8_t* const msg, size_t msg_size );
 
-/* declaration of fn pointer for the polling push mechanism */
+/* declaration of fn pointer for the publishing functions */
 typedef xi_state_t ( *on_msg_push_fn_t )(
-    const mqtt_topic_descr_t* const mqtt_topic_descr );
+    const char* mqtt_topic_descr );
 
 /* each topic will have it's descriptor and the handler in case of the message */
 typedef struct mqtt_topic_configuration_s
 {
-    mqtt_topic_descr_t topic_descr;
+    const char* const topic_name;
     on_msg_handler_fn_t on_msg_pull;
     on_msg_push_fn_t on_msg_push;
 } mqtt_topic_configuration_t;
@@ -125,27 +122,28 @@ typedef struct mqtt_topic_configuration_s
 static void on_led_msg( const uint8_t* const msg, size_t msg_size );
 
 /* declaration of topic handlers for PUB'ed topics */
-static xi_state_t pub_accelerometer( const mqtt_topic_descr_t* const mqtt_topic_descr );
-static xi_state_t pub_magnetometer( const mqtt_topic_descr_t* const mqtt_topic_descr );
-static xi_state_t pub_barometer( const mqtt_topic_descr_t* const mqtt_topic_descr );
-static xi_state_t pub_button( const mqtt_topic_descr_t* const mqtt_topic_descr );
-static xi_state_t pub_gyroscope( const mqtt_topic_descr_t* const mqtt_topic_descr );
-static xi_state_t pub_humidity( const mqtt_topic_descr_t* const mqtt_topic_descr );
-static xi_state_t pub_temperature( const mqtt_topic_descr_t* const mqtt_topic_descr );
+static xi_state_t pub_accelerometer( const char* topic_name );
+static xi_state_t pub_magnetometer( const char* topic_name );
+static xi_state_t pub_barometer( const char* topic_name );
+static xi_state_t pub_button( const char* topic_name );
+static xi_state_t pub_gyroscope( const char* topic_name );
+static xi_state_t pub_humidity( const char* topic_name );
+static xi_state_t pub_temperature( const char* topic_name );
 
 /* topic's initialisation function */
 static xi_state_t init_xively_topics( xi_context_handle_t in_context_handle );
 
-/* table of topics used for this demo */
+/* table of topics used for this demo which well be published
+ * every 5 seconds through the Xively Event System Scheduler.
+ * NOTE: Accelerometer, Gyroscope, and Button data are published more
+ * frequently by the main task
+ */
 static const mqtt_topic_configuration_t topics_array[] = {
-    {{XI_TOPIC_NAME_MANGLE( "Accelerometer" )}, NULL, pub_accelerometer},
-    {{XI_TOPIC_NAME_MANGLE( "Magnetometer" )}, NULL, pub_magnetometer},
-    {{XI_TOPIC_NAME_MANGLE( "Barometer" )}, NULL, pub_barometer},
-    {{XI_TOPIC_NAME_MANGLE( "Gyroscope" )}, NULL, pub_gyroscope},
-    {{XI_TOPIC_NAME_MANGLE( "Humidity" )}, NULL, pub_humidity},
-    {{XI_TOPIC_NAME_MANGLE( "Button" )}, NULL, pub_button},
-    {{XI_TOPIC_NAME_MANGLE( "LED" )}, &on_led_msg, NULL},
-    {{XI_TOPIC_NAME_MANGLE( "Temperature" )}, NULL, pub_temperature},
+    { XI_TOPIC_NAME_MANGLE( "Magnetometer" ), NULL, pub_magnetometer},
+    { XI_TOPIC_NAME_MANGLE( "Barometer" ), NULL, pub_barometer},
+    { XI_TOPIC_NAME_MANGLE( "Humidity" ), NULL, pub_humidity},
+    { XI_TOPIC_NAME_MANGLE( "LED" ), &on_led_msg, NULL},
+    { XI_TOPIC_NAME_MANGLE( "Temperature" ), NULL, pub_temperature},
 };
 
 /* store the length of the array */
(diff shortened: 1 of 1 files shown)


examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -90,90 +100,51 @@ char* ssid          = "SSID";
...
+static xi_state_t pub_gyroscope( const char* topic_name );
+static xi_state_t pub_humidity( const char* topic_name );
+static xi_state_t pub_temperature( const char* topic_name );
 
 /* topic's initialisation function */
 static xi_state_t init_xively_topics( xi_context_handle_t in_context_handle );
 
-/* table of topics used for this demo */
-static const mqtt_topic_configuration_t topics_array[] = {
-    {{XI_TOPIC_NAME_MANGLE( "Accelerometer" )}, NULL, pub_accelerometer},
```
atigyi: removing this you peel down an abstraction layer, why?

atigyi: Also the outgoing messages have to be known at the publish time. Having this abstraction we have to define all this config (incoming topics, outgoing topics and how to handle these) only here, and the user don't have to know the content itself.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -356,58 +331,69 @@ void on_connected( xi_context_handle_t in_context_handle, void* data, xi_state_t
...
+
+            if ( params->message.temporary_payload_data[0] == '0' )
+            {
+                io_led_off();
+            }
+            else
             {
-                topic_conf->on_msg_pull( params->message.temporary_payload_data,
-                                         params->message.temporary_payload_data_length );
+                io_led_on();
```
atigyi: here, you know that you have to call the io_led_on, since this is the only incoming topic. But otherwise how to differentiate. With an if analysing the topic name?

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -356,58 +331,69 @@ void on_connected( xi_context_handle_t in_context_handle, void* data, xi_state_t
...
-        if ( NULL != topic_config->on_msg_push )
-        {
-            topic_config->on_msg_push( &topic_config->topic_descr );
-        }
+        pub_gyroscope( XI_TOPIC_NAME_MANGLE( "Gyroscope" ) );
+        pub_accelerometer( XI_TOPIC_NAME_MANGLE( "Accelerometer" ) );
+        pub_button( XI_TOPIC_NAME_MANGLE( "Button" ) );
     }
+
+    pub_magnetometer( XI_TOPIC_NAME_MANGLE( "Magnetometer" ) );
```
atigyi: Yeah, either doing this: getting rid of abstraction allow to publish each sensor however one wants. Or extention the abstraction with a member: disable_upon_burst_mode

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -1,13 +1,21 @@
 /**
  ******************************************************************************
  * @file    main.c
- * @author  Central LAB
- * @version V2.1.0
- * @date    17-May-2016
- * @brief   Main program body
+ * @author  Xively
+ * @version V1.0.0
+ * @date    03-March-2017
```
atigyi: :+1:

atigyi:
I liked the abstraction in the topic solution and their handling function but I can accept that it is not needed here and readability is more important.
Approved :man_firefighter:

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
DellaBitta: Because it's way too complicated for a static demo.  The users has to fight their way through an abstracted table system for their "ease of use" but I don't think it's actually easy.  The table includes confusing fields including in going and outgoing data.  Isn't it easier to use pub when you want to pub in the logic and subscribe to the one topic after the connection then to conflate the two in one abstract table ? I think so.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
DellaBitta: each topic could have its own callback function.

Palantir555: I also like having different functions for SUBACK error handling and actual payload processing, even if they're both for the LED topic only.

It would also help take away some of the libxively-specific funniness (e.g. `xi_sub_call_params_t `, `params->message.temporary_payload_data_*`, etc.) by having a very simple and generic `void on_led_msg( char* message, uint len)` and a very libxively-specific `void on_led_topic_data( xi_context_handle_t [...] )`

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -356,58 +331,69 @@ void on_connected( xi_context_handle_t in_context_handle, void* data, xi_state_t
...
     {
-        const mqtt_topic_configuration_t* const topic_config = &topics_array[i];
-
-        if ( NULL != topic_config->on_msg_push )
-        {
-            topic_config->on_msg_push( &topic_config->topic_descr );
-        }
+        pub_gyroscope( XI_TOPIC_NAME_MANGLE( "Gyroscope" ) );
+        pub_accelerometer( XI_TOPIC_NAME_MANGLE( "Accelerometer" ) );
+        pub_button( XI_TOPIC_NAME_MANGLE( "Button" ) );
```
Palantir555: We need to pub_button() every time so CPM gets a `0` whenever the button stops being pressed, regardless of the state of is_burst_mode.

That is because there's no interrupt configured for the "button released" edge. I don't think it's possible to configure such interrupt for this pin in this device (there's no 'signal edge' parameter in `BSP_PB_Init()`), but I can look into it if it becomes a problem.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -417,27 +403,17 @@ xi_state_t init_xively_topics( xi_context_handle_t in_context_handle )
...
-    for ( ; i < topics_array_length; ++i )
-    {
-        const mqtt_topic_configuration_t* const topic_config = &topics_array[i];
-
-        if ( NULL != topic_config->on_msg_pull )
-        {
-            ret = xi_subscribe( in_context_handle, topic_config->topic_descr.name,
-                                XI_MQTT_QOS_AT_MOST_ONCE, &on_mqtt_topic,
-                                ( void* )topic_config );
+    ret = xi_subscribe( in_context_handle, XI_TOPIC_NAME_MANGLE( "LED" ),
```
Palantir555: :+1:

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -597,7 +571,14 @@ int main( void )
...
                 fflush( stdout );
-                HAL_Delay( 250 );
+                HAL_Delay( 100 );
+
+                if ( is_burst_mode && XI_CONNECTION_STATE_OPENED == xi_connection_state )
+                {
+                    pub_gyroscope( XI_TOPIC_NAME_MANGLE( "Gyroscope" ) );
+                    pub_accelerometer( XI_TOPIC_NAME_MANGLE( "Accelerometer" ) );
+                    pub_button( XI_TOPIC_NAME_MANGLE( "Button" ) );
+                }
```
Palantir555: :+1:

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -82,6 +90,8 @@ WiFi_Status_t wifi_get_AP_settings( void );
 wifi_state_t wifi_state;
 wifi_config config;
 wifi_scan net_scan[WIFI_SCAN_BUFFER_LIST];
+xi_state_t xi_connection_state = XI_SOCKET_NO_ACTIVE_CONNECTION_ERROR;
```
DellaBitta: That's a good idea.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -90,90 +100,51 @@ char* ssid          = "SSID";
...
+typedef xi_state_t ( *on_msg_push_fn_t )( const char* mqtt_topic_descr );
 
 /* each topic will have it's descriptor and the handler in case of the message */
 typedef struct mqtt_topic_configuration_s
 {
-    mqtt_topic_descr_t topic_descr;
+    const char* const topic_name;
     on_msg_handler_fn_t on_msg_pull;
     on_msg_push_fn_t on_msg_push;
 } mqtt_topic_configuration_t;
```
DellaBitta: Good call!

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -617,6 +598,16 @@ void HAL_GPIO_EXTI_Callback( uint16_t GPIO_Pin )
 {
     if ( GPIO_Pin == IO_NUCLEO_BUTTON_PIN )
     {
+        is_burst_mode = !( is_burst_mode );
```
DellaBitta: Done.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -842,24 +833,22 @@ ex: printf("Wrong parameters value: file %s on line %d\r\n", file, line) */
...
     uint8_t console_input[1], console_count = 0;
-    //  wifi_bool set_AP_config = WIFI_FALSE;
     WiFi_Status_t status = WiFi_MODULE_SUCCESS;
     printf( "\r\n\n/********************************************************\n" );
-    printf( "\r *                                                      *\n" );
-    printf( "\r * X-CUBE-WIFI1 Expansion Software v2.1.1               *\n" );
-    printf( "\r * X-NUCLEO-IDW01M1 Wi-Fi Configuration.                *\n" );
-    printf( "\r * Client-Socket Example                                *\n" );
-    printf( "\r *                                                      *\n" );
+    printf( "\r *\n" );
```
DellaBitta: cool, yeah they were pretty clunky.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
DellaBitta: And yes, I wasn't sure if I should add it to the button or to make a new subscription topic that can force the device to enter burst mode from the Xively backend. 

Ultimately I thought that the latter would make more sense, but decided on the button for now due to time constraints.

PR #106, opened 2017-03-06
User data storage in flash and runtime user configuration via UART

Palantir555:
[DESCRIPTION]
The user settings for the demo are now configurable at runtime. A datastructure is used to keep them all together: `user_data_t`. That structure contains static buffers for the **wifi SSID**, **wifi password**, **wifi encryption mode**, **xively account ID**, **xively device ID** and **xively device password**, as well as a **CRC checksum** of all the previous data, used to validate the integrity of the data we retrieve from flash on boot.

The user data storage interface is independent from the provisioning process. A proof of concept is provided using the UART debug port to set the credentials, in the hopes we can replace it with a better solution (HTTP || NFC) in the short term. Hardcoded credentials, and runtime configuration with no flash storage can both be selected via macros in main.c

In order to create the data storage PoC, main.c had to be heavily refactored so the MQTT topics can be mangled at runtime. As part of that refactoring, I took all the low level peripheral configuration (clock settings, etc.) to a new `demo_bsp.c` file so they don't clutter main.c.

Finally, this branch normalises the UART logs and removes some unnecessary debug messages, resulting in much more readable output.

WARNING: This PR uses a very small chunk of data in the last sector in the device's flash memory. The Linker script has not been modified not to write code into this sector (for portability reasons), and the source code is not ready to read non-user_data before erasing the sector and re-writing it all (but the user data segment) after the fact.

[REVIEWERS]

[JIRA]
Implement user data storage in flash so the device can be re-configured without re-flashing. [Task: ??]

[QA]
1. Store valid data to flash with the correct checksum, reboot the device. It successfully retrieves the data, validates it and connects to the WiFi network and Xively
2. Store valid data to flash with an incorrect checksum, reboot the device. It successfully retrieves the data, its validation fails and the user is prompted for new credentials
3. Store valid credentials with valid checksum in flash, reboot the device while keeping the user button pressed. The user is asked prompted for new credentials regardless of the state of the ones in flash.
4. In configuration mode, set wifi credentials but not xively credentials and validate they're correctly stored and used. Do the same thing but modifying the xively credentials and not the wifi ones.

[RELEASE NOTES]
N/A


Palantir555:
NOTE: There's something broken in the usage of the new topic mangler implementation in main.c. It crashes the program after successfully connecting and subscribing, probably when it tries to publish. I'm looking into it.

Palantir555:
I'll be solving the new conflicts shortly. PR not ready for review yet.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Inc/user_data.h
```diff
@@ -0,0 +1,39 @@
+#ifndef __USER_DATA_H__
+#define __USER_DATA_H__
+
+#define USER_DATA_STR_LEN 64 /* Must be a multiple of sizeof(int32_t) for CRC periph */
+
+typedef struct
+{
+    int32_t crc_checksum; /* For internal use only - Set by user_data_save_to_flash() */
+
+    int32_t wifi_client_encryption_mode; /* Casted from WiFi_Priv_Mode */
```
atigyi: yeah, this is wifi specific, the ethernet version won't need this. we'll have to handle this.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/demo_bsp.c
```diff
@@ -0,0 +1,196 @@
...
+    if ( HAL_RCC_ClockConfig( &clkinitstruct, FLASH_LATENCY_2 ) != HAL_OK )
+    {
+        /* Initialization Error */
+        while ( 1 )
+            ;
+    }
+}
+#endif
+
+#ifdef USE_STM32F4XX_NUCLEO
```
atigyi: here use #elif defined(ABC) here since these are exclusive codes.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -78,180 +80,298 @@
...
+#define ACCELEROMETER_CHANNEL_NAME "Accelerometer"
+#define GYROSCOPE_CHANNEL_NAME     "Gyroscope"
+#define MAGNETOMETER_CHANNEL_NAME  "Magnetometer"
+#define BAROMETER_CHANNEL_NAME     "Barometer"
+#define HUMIDITY_CHANNEL_NAME      "Humidity"
+#define TEMPERATURE_CHANNEL_NAME   "Temperature"
+#define BUTTON_CHANNEL_NAME        "Button"
+#define LED_CHANNEL_NAME           "LED"
+
+/* RAM can be saved by building each of these when needed and free()ing them after use */
```
atigyi: so this is a RAM gain for SPEED cost. Are we in RAM pressure?

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Inc/user_data.h
Palantir555: Well, if we don't set the `wifi_` fields, they'll be saved as 0s and everything will work as expected. We'd be wasting a couple hundred bytes, and we'd have to explain in the Eth source code that some variables are not necessary, but we'd also save some implementation/maintenance trouble. Do you think it's worth the effort of maintaining 2 different user_data interfaces?

If we decide to remove them it won't be a big deal.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
Palantir555: We're not, that's why we're not free()ing those strings. I just added the comment in case a user used the example as a template for a device with a smaller footprint, but I'm happy to remove it if you think it's confusing.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
```diff
@@ -0,0 +1,342 @@
...
+
+/**
+ * @brief  Copy the Flash area reserved for user data into *dst so it can be
+ *         manipulated
+ * @param  *dst is the pre-allocated pointer where we want to copy the user data
+ * @retval <0: Error, >=0: OK
+ */
+int8_t user_data_copy_from_flash( user_data_t* dst )
+{
+    memset( dst, 0x00, sizeof( user_data_t ) );
```
atigyi: pls check dst != NULL here

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
```diff
@@ -0,0 +1,342 @@
...
+#include "stm32f4xx.h"
+#include "stm32f4xx_hal_conf.h"
+#include "user_data.h"
+
+#define CRC_REGISTER_BYTES 4
+static int8_t calculate_checksum( user_data_t* user_data, int32_t* out_checksum );
+
+#define FLASH_USER_DATA_SECTOR FLASH_SECTOR_TOTAL-1
+#define FLASH_VOLTAGE_RANGE    FLASH_VOLTAGE_RANGE_1
+#define FLASH_USER_DATA_PREFIX 0x55
```
atigyi: is this FLASH_USER_DATA_PREFIX used at all somewhere? I didn't find it.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
```diff
@@ -0,0 +1,342 @@
...
+ */
+#include <stdlib.h>
+#include <string.h>
+#include <assert.h>
+#include <stdint.h>
+#include "stm32f4xx.h"
+#include "stm32f4xx_hal_conf.h"
+#include "user_data.h"
+
+#define CRC_REGISTER_BYTES 4
```
atigyi: I'd define this with sizeof( uint32_t ) or at least write comment why 4

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -78,180 +80,298 @@
...
         return -1;
     }
-    if ( io_axes_to_json( sensor_input, out_msg, IO_AXES_JSON_BUFFER_MAX_SIZE ) < 0 )
+    if ( 0 > io_axes_to_json( sensor_input, out_msg, IO_AXES_JSON_BUFFER_MAX_SIZE ) )
     {
-        printf( "\r\n\t[ERROR] trying to create JSON string from sensor input" );
         return -1;
     }
-    return xi_publish( gXivelyContextHandle, topic_name, out_msg,
+    return xi_publish( gXivelyContextHandle, mqtt_topics.magnetometer, out_msg,
```
atigyi: here I don't see the benefit storing all the topic strings in a struct variable instead of using the previous macro: XI_TOPIC_NAME_MANGLE( "Magnetometer" ). Is there benefit?

atigyi: Is one better than the other?

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -808,87 +753,197 @@ ex: printf("Wrong parameters value: file %s on line %d\r\n", file, line) */
...
+            case 0:
                 break;
             default:
-                printf( "\r\nWrong Entry. Priv Mode is not compatible\n" );
-                return WiFi_NOT_SUPPORTED;
+                printf( "\r\n\t[ERROR] Getting AP credentials from user" );
+                return -1;
         }
+
+        printf( "\r\n>> Saving user data to flash" );
```
atigyi: is this save needed?

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
Palantir555: It's not (anymore). I'll get rid of it

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
Palantir555: XI_TOPIC_NAME_MANGLE is run at compile time; we don't have the Account ID and Device ID at compile time anymore, so it's not an option.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
Palantir555: It's not -- Good catch; I'll remove it.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -424,39 +526,68 @@ int main( void )
...
+        printf( "\r\n>> User data flash initialization [ERROR]" );
+        return -1;
+    }
+#endif
+
+    /* Init the sensor board */
     printf( "\r\n>> Initializing the sensor extension board" );
     io_nucleoboard_init();
     io_sensorboard_init();
     io_sensorboard_enable();
```
Palantir555: Note to myself: `return -1;` if any of these calls returns `<0`.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Inc/user_data.h
atigyi: It is misleading. I'd be confused to read anything about the wifi in an ethernet project. Especially if there are two types: wifi and ethernet. I'd be confused where I am. But yeah, we may put there a comment and that's it. We have to handle this keeping the weight of this project in mind. So it is ok for now.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
atigyi: :+1:

Palantir555:
@atigyi I've updated the contents of this PR with your suggestions.

I've also looked into removing the WiFi credentials from user_data. We only need to remove a few lines from `user_data.h` (the wifi_ fields from the datastructure and their setters), and update the calculate_checksum() function.
It should be very straightforward; let's port it to the ethernet project as it is, and I can remove it later.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
```diff
@@ -0,0 +1,361 @@
...
+
+/**
+ * @brief  This function calculates and sets the CRC checksum of the user data
+ *         we'd like to save, erases the relevant Fash sector and stores the new
+ *         data
+ * @param  *src is the datastructure we'd like to save to flash. src->crc_checksum
+ *         will be overwritten
+ * @retval <0: Error, >=0: OK
+ */
+/* TODO: This execution branch should run entirely from volatile RAM if possible */
```
DellaBitta: should this TODO still be here?

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
```diff
@@ -0,0 +1,361 @@
...
+int8_t user_data_flash_init( void )
+{
+    assert( 0 == sizeof( user_data_t ) % sizeof( int32_t ) ); /* For program_flash() */
+    assert( 0 == USER_DATA_STR_SIZE % CRC_REGISTER_SIZE );
+    if ( !IS_FLASH_SECTOR( FLASH_USER_DATA_SECTOR ) )
+    {
+        printf( "\r\n>> Flash init [ERROR] Sector [%d] is incompatible with this device",
+                FLASH_USER_DATA_SECTOR );
+        return -1;
+    }
```
DellaBitta: New line.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
```diff
@@ -0,0 +1,361 @@
...
+    if ( calculate_checksum( src, &src->crc_checksum ) < 0 )
+    {
+        printf( "\r\n\tCRC checksum calculation [ERROR]" );
+        return -1;
+    }
+
+    if ( user_data_reset_flash() < 0 )
+    {
+        printf( "\r\n\tSaving user data to flash [ABORT]" );
+        return -1;
```
DellaBitta: Is there a reason why the new lines are prepended and not at the end of the printf() string?

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
```diff
@@ -0,0 +1,361 @@
...
+}
+
+/**
+ * @brief  Performs a Flash erase on the section reserved for user data
+ * @retval <0: Error, >=0: OK
+ */
+/* TODO: This execution branch should run entirely from volatile RAM if possible */
+int8_t user_data_reset_flash( void )
+{
+    /* TODO: Save all non-user_data in the sector before the erase */
```
DellaBitta: I'm not sure we're coming back to this code any time soon, therefore I don't know if we should have TODOs in here.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
```diff
@@ -0,0 +1,361 @@
...
+        }
+        printf( "%08lx ", *( ( int32_t* )user_data + i ) );
+    }
+#endif
+    printf( "\r\n\t* WiFi Security: [%ld]", user_data->wifi_client_encryption_mode );
+    printf( "\r\n\t* WiFi SSID: [%.64s]", user_data->wifi_client_ssid );
+    printf( "\r\n\t* WiFi Pwd: [%.64s]", user_data->wifi_client_password );
+    printf( "\r\n\t* Xi Acc ID: [%.64s]", user_data->xi_account_id );
+    printf( "\r\n\t* Xi Dev ID: [%.64s]", user_data->xi_device_id );
+    printf( "\r\n\t* Xi Dev Pwd: [%.64s]", user_data->xi_device_password );
```
DellaBitta: I think we shouldn't print out the xively device password to serial.  It would comment out this line and add a note saying that they COULD enable this if they need to debug something, but it's suppressed for security reasons.

Perhaps also fro the Wifi PWD, too.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
Palantir555: I think if we're gonna publish code that might cause critical errors on other devices, we should document its weaknesses. It tells the user how to implement it properly if they use it for their device.

Just my 2 cents, though. Let me know what you think.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
Palantir555: At first I was just following the convention in the rest of the application. Now I kinda like having the `\r\n` together with the `\t`s, as well as not bothering to check whether the previous print added the `\n` or not (e.g. loops printing `.`).

I don't mind changing it, though. Is there a libxively-wide convention?

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
Palantir555: See https://github.com/xively/xively-client-c/pull/106/files/2329aa6dc90c31da480da58ab5f0d539078d82fd#r107500026

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
DellaBitta: It is, but this isn't technically libxively.  I've had to follow some TI conventions in their example source, otherwise our adaptations stick out like a sore thumb.  :'/  

So, Ok I guess. :)

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
DellaBitta: Hmm.. not sure if it's a TODO then, right?  Todo usually means "We'll revisit this when we have time"

DellaBitta: Seems more like a WARNING or NOTE

Palantir555: Well, I'd love to revisit this if we have the time. I think it's an implementation with good business value (in a production environment it would be used to store multiple datastructures in a memory area reserved at the linker script, not to separate code from user data, but the code is the same).

That being said, I can certainly change the comments to `WARNING`s if we're not gonna have the time to implement this.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
Palantir555: Good point :+1:

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/user_data.c
DellaBitta: Yeah, I would put a Warning or a Note, and at the top of the source, in the header, discuss potential future improvements. 

TODOs, to me, make like it seem that it's unfinished and not really tested / complete, which isn't really something that breeds confidence in new evaluators, especially if it works fine for this intended platform.

If it DOESN'T then we need to make a jira story so that it does.

PR #118, opened 2017-06-02
Feature/secure file transfer

atigyi:
[ Description ]
This is the first chunk of Secure File Transfer + Firmware Update support inside libxively. This includes the CBOR codec and some skeleton of the SFT logic inside the Xively C Client architecture. The code here is unittested and and a first step is there for integrations test of SFT messaging between Client and Broker.
This change also includes a Crypt BSP module for SHA256 although this may not remain since the Fingerprint used in SFT is not necessarily SHA256 can be any checksum. Whether this remains as separate BSP module or removed will turn out later in the dev process.

[ JIRA ]
SFT implementation inside libxively
https://jira.3amlabs.net/browse/XCL-2938

[ Reviewer ]
@DellaBitta

[ QA ]
Unit tests on CBOR codec.

[ Release Notes ]


make/mt-config/mt-config.mk
```diff
@@ -78,6 +78,8 @@ XI_CONFIG_FLAGS += -DXI_DEBUG_OUTPUT=$(XI_DEBUG_OUTPUT)
 XI_CONFIG_FLAGS += -DXI_DEBUG_ASSERT=$(XI_DEBUG_ASSERT)
 XI_CONFIG_FLAGS += -DXI_DEBUG_EXTRA_INFO=$(XI_DEBUG_EXTRA_INFO)
 
+XI_CONFIG_FLAGS += -DUSE_CBOR_CONTEXT
```
atigyi: Enabling CBOR context enables us to inject memory handler functions (alloc, calloc). Thus this helps us to create leak free code.

src/libxively/xi_globals.h
```diff
@@ -33,6 +33,9 @@ typedef struct
     char* str_account_id;
     char* str_device_unique_id;
     xi_backoff_status_t backoff_status;
+
+    char** updateable_files;
+    uint16_t updateable_files_count;
```
DellaBitta: shouldn't this be in the context?  What if two connections are built, and they both get file update available messages?  Wont' they conflict with each other?

src/tests/itests/xi_itest_update.c
```diff
@@ -0,0 +1,91 @@
...
+    /* clear the external dependencies */
+    xi_context            = NULL;
+    xi_context_handle     = XI_INVALID_CONTEXT_HANDLE;
+    xi_context_mockbroker = NULL;
+
+    xi_memory_limiter_tearup();
+
+    *fixture_void = NULL;
+    // *fixture_void = xi_itest_connect_error__generate_fixture();
+
```
DellaBitta: Are these commented out lines applicable anymore?

src/tests/itests/xi_itest_update.c
```diff
@@ -0,0 +1,91 @@
...
+    xi_globals.backoff_status.backoff_lut_i = 0;
+    xi_cancel_backoff_event();
+
+    xi_initialize( "xi_itest_connect_error_account_id",
+                   "xi_itest_connect_error_device_id" );
+
+    // xi_initialize_add_updateable_file( ( const char* [] ){"file1", "file2", "file3"}, 3
+    // );
+    // xi_initialize_add_updateable_file( ( const char* [] ){"file1"}, 1 );
+
```
DellaBitta: and here?

src/tests/itests/xi_itests.c
```diff
@@ -12,20 +12,21 @@
...
-                                                      ,
-    cmocka_test_group( xi_itests_tls_layer )
+                               cmocka_test_group( xi_itests_tls_layer ),
 #endif
-        ,
-    cmocka_test_group( xi_itests_mqttlogic_layer ),
-    cmocka_test_group( xi_itests_connect_error ), cmocka_test_group_end};
+                               cmocka_test_group( xi_itests_mqttlogic_layer ),
+                               cmocka_test_group( xi_itests_connect_error ),
+                               cmocka_test_group( xi_itests_update ),
```
DellaBitta: should this be named sft or something more feature-specific than update?  Update what?  It's not really saying. :(

include/bsp/xi_bsp_crypt.h
```diff
@@ -0,0 +1,16 @@
...
+#ifndef __XI_BSP_CRYPT_H__
+#define __XI_BSP_CRYPT_H__
+
+#include <stdint.h>
+
+int xi_bsp_crypt_sha256_init( void** );
+int xi_bsp_crypt_sha256_update( void*, const uint8_t*, uint32_t );
+int xi_bsp_crypt_sha256_final( void*, uint8_t* );
+
+#endif // __XI_BSP_CRYPT_H__
```
atigyi: done

make/mt-config/mt-config.mk
```diff
@@ -186,6 +191,10 @@ XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/datastructures
 XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/mqtt/codec
 XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/mqtt/logic
 XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/control_topic
+XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/cbor
+XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/cbor/cn-cbor/src
+# XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/cbor/tinycbor/src
```
atigyi: "TinyCBOR is Intel's industrial strength C/C++ implementation of CBOR, as used in the IoTivity framework:"
http://cbor.io/impls.html

I tried another cbor impl because the cn-cbor had some porting issues. But turned out that the other impls have porting issues as well. So I chose cn-cbor (Alan's choice).

And good catch, I remove this line from makefile!

include/bsp/xi_bsp_crypt.h
atigyi: also searched for "#endif // __XI" and replaced them with /* */.

make/mt-os/mt-stm32fx.mk
```diff
@@ -90,6 +90,7 @@ endif
 XI_CONFIG_FLAGS += -DXI_CROSS_TARGET
 XI_CONFIG_FLAGS += -DXI_EMBEDDED_TESTS
 
+XI_COMPILER_FLAGS += -DLWIP_TIMEVAL_PRIVATE=0
```
atigyi: very good catch! This is not and SFT issue. This was an old issue back when I did the port on STM32 Eth which uses LWIP as network stack. timeval is defined in stdio.h and in LWIP somewhere (maybe netdb.h ?). And it conflits. Which resulted in the inability of emitting debug output in code which is compiled against STM32 Eth. And this has not only effect on STM32 but common code as well. And I faced with this problem too: When I left accidentally (or intentionally) a printf in a common code then the STM32 eth build failed with: conflicting defines on timeval.

And no I found that timeval def can be tuned of in LWIP. So this is a "side" maintenance change.

src/bsp/platform/cc3200/xi_bsp_hton.h
```diff
@@ -0,0 +1,2 @@
+#include <socket.h>
+typedef long int ssize_t;
```
atigyi: Ok, so this is the solution of one porting issue the cn-cbor has. cn-cbor needs hton function. And apparently it needs ssize_t too. hton is full platform dependent, architecture dependent? It is about the endianness. So to have cn-cbor on all of our target platforms we need hton which is platform dependent --> new BSP module without API (.h).

So this file is called "hton" but it isn't just hton but ssize_t too. And its actual job is to port cn-cbor. So it could be named "xi_bsp_cncbor_port.c". But it may help others who need hton. So I called it xi_bsp_hton.c but ssize_t appeared. So the best name could be: xi_bsp_hton_and_ssize_t.c :) Or xi_bsp_some_stuff.c :) Or we should break out from this BSP world representation and figure out the ultimate solution, or not ultimate but better than this.
So this is the tale of xi_bsp_hton.c

src/bsp/platform/stm32fx_nucleo_wifi/xi_bsp_hton.h
```diff
@@ -0,0 +1,51 @@
+#ifndef __XI_BSP_HTON_H__
```
atigyi: It is. It is a networking function converting endianness between network byte order and host byte order. The thing is the WIFI. The STM32 WIFI SDK completely hides networking. We don't have access to hton low level function. Thus cn-cbor is in trouble on STM32 WIFI. That is why the implementation is here :(

src/bsp/tls/crypto-algorithms/README
```diff
@@ -0,0 +1,3 @@
+Files are from Brad Conte at: https://github.com/B-Con/crypto-algorithms
+Files are public domain
+Files have been slightly modified to adjust necessary #include files for compilation
```
atigyi: :+1: but according to the page https://github.com/B-Con/crypto-algorithms I think is ok leaving this.

Especially because according to the conversation with the Platform Team (GTyukas, JKlemm, LKorossy) the actual implementation of FINGERPRINT will be per customer / use case. It may be any checksum (MD5, sha256, CRC code, ...). Which means it won't be part of the Xively C Client but the Application. Probably this will be another BSP module: xi_bsp_update_fingerprint.h or just xi_bsp_update.h (I would like to rename the xi_bsp_sft.h to xi_bsp_update.h if you also agree).

src/bsp/tls/crypto-algorithms/xi_bsp_sha256.c
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * This is part of the Xively C Client library,
+ * it is licensed under the BSD 3-Clause license.
+ */
+
+#include <xi_bsp_crypt.h>
+#include <xi_bsp_mem.h>
+#include <sha256.h>
+
+int xi_bsp_crypt_sha256_init( void** sha )
```
atigyi: done

src/bsp/tls/crypto-algorithms/xi_bsp_sha256.c
```diff
@@ -0,0 +1,34 @@
...
+{
+    *sha = xi_bsp_mem_alloc( sizeof( SHA256_CTX ) );
+
+    sha256_init( ( SHA256_CTX* )*sha );
+
+    return 0;
```
atigyi: 1: I think the SHA BSP module will be deleted. It will be the application's responsibility to provide a fingerprint service to a file. Although the Xively C Client may provide a "default" implementation if one doesn't want to spend much time on this, especially in case of a tutorial. Or an SHA256 may be enough for 90% of customers. So we should provide a default fingerprint service somehow, maybe no in the BSP module format. This will turn out during the fingerprint "redesign" since until now I thought it will be always SHA256.
There will be a BSP fingerprint module which the application can implement and this implementation can call back to a new Xively C Client API: xi_sha256. Which has implementations internally, pointing to wolfSSL / mbedTLS / Brad Conte's sha256 impl.
2. Unfortunately Brad Conte's functions return void.

src/bsp/tls/mbedtls/xi_bsp_sha256.c
```diff
@@ -0,0 +1,36 @@
...
+    *sha = xi_bsp_mem_alloc( sizeof( mbedtls_sha256_context ) );
+
+    mbedtls_sha256_init( ( mbedtls_sha256_context* )*sha );
+
+    return 0;
+}
+
+int xi_bsp_crypt_sha256_update( void* sha, const uint8_t* data, uint32_t len )
+{
+    mbedtls_sha256_starts( ( mbedtls_sha256_context* )sha, 0 );
```
atigyi: it is a flag mbedTLS offers, it is described in the API:

/**
 * \brief          SHA-256 context setup
 *
 * \param ctx      context to be initialized
 * \param is224    0 = use SHA256, 1 = use SHA224
 */
void mbedtls_sha256_starts( mbedtls_sha256_context *ctx, int is224 );

atigyi: I don't think we should have function descriptions at call places. The documentation is at the definition.

src/libxively/cbor/cn-cbor/src/cn-encoder.c
```diff
@@ -0,0 +1,309 @@
...
+    ensure_writable(9);
+    be64 = hton64p((const uint8_t*)&u64.u);
+
+    write_byte_and_data(IB_PRIM | 27, (const void*)&be64, 8);
+
+  }
+}
+#endif /* CBOR_NO_FLOAT */
+
+// TODO: make public?
```
atigyi: :smile: it is from CBOR. I think I should handle external library imports in separate pull requests.

src/libxively/cbor/xi_cbor_codec_ct.c
```diff
@@ -0,0 +1,344 @@
...
+void xi_cbor_put_name_and_revision( cn_cbor* cb_map,
+                                    const char* name,
+                                    const char* revision,
+                                    cn_cbor_errback* errp )
+{
+    if ( NULL != name )
+    {
+        cn_cbor_map_put( cb_map,
+                         cn_cbor_string_create(
+                             XI_CBOR_CODEC_CT_STRING_FILE_NAME CBOR_CONTEXT_PARAM, errp ),
```
atigyi: :) error pointer :)

src/libxively/cbor/xi_cbor_codec_ct.c
```diff
@@ -0,0 +1,344 @@
...
+#define XI_CBOR_CODEC_CT_STRING_FILE_FINGERPRINT "F"
+
+#define XI_CBOR_CODEC_CT_STRING_FILECHUNK_OFFSET "O"
+#define XI_CBOR_CODEC_CT_STRING_FILECHUNK_LENGTH "L"
+
+void xi_cbor_put_name_and_revision( cn_cbor* cb_map,
+                                    const char* name,
+                                    const char* revision,
+                                    cn_cbor_errback* errp )
+{
```
atigyi: This is an internally used function and NULL check is done later. I think quality is enough here. In general on this level we may allow messages without name and/or revision. Although is isn't used without name and revision. So I'd leave this as it is for now if it is ok for you.

src/libxively/cbor/xi_cbor_codec_ct.c
```diff
@@ -0,0 +1,344 @@
...
+        cn_cbor_int_create( control_message->common.msgver CBOR_CONTEXT_PARAM, &err ),
+        &err );
+
+    switch ( control_message->common.msgtype )
+    {
+        case XI_CONTROL_MESSAGE_DB_FILE_INFO:
+
+            if ( 0 < control_message->file_info.list_len )
+            {
+                cn_cbor* files = cn_cbor_array_create( CBOR_CONTEXT_PARAM_COMA & err );
```
atigyi: hm, unfortunately the code formatter does this because of the macro: CBOR_CONTEXT_PARAM_COMA. The formatter does not expand macros.

atigyi: I don't think clang-format can be setup to handle this :(

src/libxively/control_topic/xi_control_message.c
```diff
@@ -0,0 +1,115 @@
...
+
+err_handling:
+
+    return file_info;
+}
+
+void xi_control_message_free( xi_control_message_t** control_message )
+{
+    if ( NULL == control_message || NULL == *control_message )
+    {
```
atigyi: Assert would crash the code, I wouldn;'t use it here. I think calling a free on a NULL pointer should be ok. Calling free multiple times on a pointer, I'd accept this.

src/libxively/control_topic/xi_control_message.c
```diff
@@ -0,0 +1,115 @@
...
+        return;
+    }
+
+    switch ( ( *control_message )->common.msgtype )
+    {
+        case XI_CONTROL_MESSAGE_DB_FILE_INFO:
```
atigyi: ah, ok, renaming

atigyi: done
XI_CONTROL_MESSAGE_CS_FILE_INFO

src/libxively/control_topic/xi_control_topic_layer.c
```diff
@@ -78,22 +76,49 @@ xi_control_topic_subscribe( void* context, char* subscribe_control_topic_name );
...
 {
+    // printf( "%s, control_message: %p\n", __FUNCTION__, control_message );
+
     xi_state_t local_state = XI_STATE_OK;
 
     xi_control_topic_layer_data_t* layer_data =
         ( xi_control_topic_layer_data_t* )XI_THIS_LAYER( context )->user_data;
 
-    assert( layer_data != NULL );
+    assert( NULL != layer_data /* && NULL != data */ );
```
atigyi: :+1: done

src/libxively/control_topic/xi_control_topic_layer.c
```diff
@@ -78,22 +76,49 @@ xi_control_topic_subscribe( void* context, char* subscribe_control_topic_name );
...
+    if ( NULL == control_message )
+    {
+        return XI_INVALID_PARAMETER;
+    }
```
atigyi: error handling. This was a topic we disputed about with Olgierd. Olgierd likes assertion. I like it less. Here I wouldn't crash the program just because somebody wants to send a NULL control message. I think this situation can be handled: it is a no-operation.
What I did is I moved this condition to be the first action in the function. Hope this excuse is ok for you :) (No is a good answer here :) )

src/libxively/sft/xi_sft_logic.c
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+#include <xi_sft_logic.h>
+#include <stddef.h>
+#include <xi_macros.h>
+
+xi_state_t xi_sft_make_context( xi_sft_context_t** context,
+                                const char** updateable_files,
+                                uint16_t updateable_files_count,
+                                fn_send_control_message fn_send_message,
```
atigyi: exactly. This is the decoupling. With this move the xi_sft_logic.c does not need to include the xi_control_topic_layer.h. Saying it one level higher: SFT does not know about Control Topic Layer, one level more: SFT does not know about transport layer? Maybe this is true. It just knows the protocol / logic.
I would have done this same way (decoupled) if there isn's a requirement to inject this SFT into other clients. Benefit of decoupling is the ease of unit testing. Thus the SFT can be tested without the control topic :) Lots of good things come from decoupling.

src/libxively/sft/xi_sft_logic.c
```diff
@@ -0,0 +1,98 @@
...
+{
+    if ( NULL == context || NULL == sft_message )
+    {
+        return XI_INVALID_PARAMETER;
+    }
+
+    switch ( sft_message->common.msgtype )
+    {
+        case XI_CONTROL_MESSAGE_BD_FILE_UPDATE_AVAILABLE:
+            // - check whether device is ready to start download of file
```
atigyi: :+1:

src/libxively/sft/xi_sft_logic.c
```diff
@@ -0,0 +1,98 @@
...
+            // - get the first chunk of file in quiestion with FILE_GET_CHUNK
```
atigyi: :+1:

src/libxively/xi_globals.h
atigyi: hm, unfortunately you are right!
ok, now there is a new API for file updates:
```
extern xi_state_t
xi_initialize_add_updateable_files( const char* filenames[], uint16_t count );
```

I will remove this and add a new with context:

```
extern xi_state_t
xi_set_updateable_files( xi_context_handle_t context_handle, const char* filenames[], uint16_t count );
```

Is is ok?

src/tests/itests/xi_itest_update.c
atigyi: This test does nothing now. This is the point I decided to put this to review since it grew too big. I will continue from here. This will change soon but for now I'd leave as it is if possible.

src/tests/itests/xi_itest_update.c
atigyi: same as above, work in progress here.

src/tests/itests/xi_itests.c
atigyi: Agree.
- device_update?
- file_update?
- firmware_update isn't good since it is not "just" firmware update.
- sft isn't enough either, since it is not only about secure file transfer, it is about maintaining the files and even applying an action on the file / device.

src/bsp/platform/cc3200/xi_bsp_hton.h
DellaBitta: hrm.. I don't think I have anything that comes to mind as an elegant way to solve this problem. :(

src/bsp/tls/crypto-algorithms/README
DellaBitta: Slightly confused. so this is going away then?

src/libxively/cbor/cn-cbor/src/cn-encoder.c
DellaBitta: Gotcha.

src/libxively/control_topic/xi_control_message.c
DellaBitta: ok

src/libxively/sft/xi_sft_logic.c
DellaBitta: OK, I think it would be nice to be able to use SFT without our client, but one thing at a time. Too many implications to wrap our heads around.

src/tests/itests/xi_itests.c
DellaBitta: I would stick with SFT since that's what it's called on the backend.  I know that there's more than just file transfer, including revision tracking, but the feature really is for SFT and that's what we're implementing.

DellaBitta: well I guess YOU'RE implementing.

src/bsp/tls/crypto-algorithms/README
atigyi: This will go away but not now. I would like to merge this to development branch. I think we don't need to add any further license to our list.

src/tests/itests/xi_itests.c
atigyi: ok, I can accept this, I will stick with SFT then, thx for the help and explanation

src/libxively/cbor/xi_cbor_codec_ct.c
```diff
@@ -0,0 +1,344 @@
...
+
+#ifdef USE_CBOR_CONTEXT
+
+#include <xi_allocator.h>
+
+void* xi_cn_calloc_func_wrapper( size_t count, size_t size, void* context )
+{
+    ( void )context;
+
+    // printf( "hello calloc wrapper: %lu, %lu\n", count, size );
```
atigyi: done

src/libxively/xi_globals.h
atigyi: moved

src/bsp/tls/crypto-algorithms/xi_bsp_sha256.c
DellaBitta: OK, so yes, I forsee this living in a particular BSP implementation for a platform / product, but not in the BSP declarations and SDK that we provide. 

Also, don't we have a BSP Success error code that we could return instead of the magic number of 0?

atigyi:
@DellaBitta I've removed the return values of the BSP SHA256 functions since 2 of 3 implementation don't return anything. What about this solution? :)

DellaBitta:
Good to go.

PR #124, opened 2017-06-23
UART provisioning and Flash storage of Xi credentials in stm32f4 ethernet demo

Palantir555:
[ Description ]
Based on the existing WiFi demo code:
- stm32f4 can now get the MQTT credentials at runtime via UART
- The credentials are stored to the last sector of flash
    - NOTE: I haven't updated this linker script yet, so it will break very large application binaries. The full explanation can be found in the header of `user_data.c`
- I've done some refactoring in the WiFi project to prepare the path towards having `Common/[Src|Inc]` folders soon:
    - Provisioning code (uart input) has been moved to a `provisioning.c` file
    - The user_data interface now uses a macro to set whether we need to allocate space for WiFi credentials or not
- `syscalls.c` has been modified, and the WiFi's `console.c` has been imported in order to support `scanf` on the stm32f429zi
- Note added to stm32f2 example explaining it's currently broken (latest SDK) and we don't plan on supporting it anymore, but it's left in there in case anybody finds it useful

[ Reviewer ]
@atigyi 

[ QA ]
- Ethernet project:
    - Build configuration `IKS01A1_Debug`: Builds, runs, re-flashes credentials OK
    - Build configuration `IKS01A1_Debug`: Builds, runs OK
- WiFi project:
    - Build configuration `IKS01A1_Debug`: Builds, runs, re-flashes credentials OK
    - Build configuration `IKS01A2_Debug`: Builds, runs OK

[ Release Notes ]
STM32F4 ethernet example now supports runtime provisioning and flash storage of Xively credentials.


examples/stm32/xively_demo_stm32f2xx_nucleo_eth/readme.md
```diff
@@ -0,0 +1,14 @@
...
+
+This example is broken for the latest version of the Cube SDK. It was proven
+working with version `STM32Cube_FW_F4_V1.14.0`, but broke at
+`STM32Cube_FW_F4_V1.16.0`
+
+We don't plan on maintaining the example for every SDK release. We're not
+removing it either, in case it's useful to developers looking to use this
+platform.
+
+You should expect a significant amount of work to get it working. An SDK update
```
atigyi: I wouldn't guess how much work is needed to get it working. Maybe it is super easy for a STM expert, don't know.

examples/stm32/xively_demo_stm32f4xx_nucleo_eth/Src/main.c
```diff
@@ -68,13 +68,6 @@ struct netif gnetif; /* network interface structure */
 UART_HandleTypeDef UartHandle;
 
 /* Private function prototypes -----------------------------------------------*/
-#ifdef __GNUC__
```
atigyi: if it is true, then big 👍 for removing this

examples/stm32/xively_demo_stm32f4xx_nucleo_eth/SW4STM32/syscalls.c
```diff
@@ -133,14 +133,9 @@ int _lseek(int file, int ptr, int dir)
...
-	int DataIdx;
-
-	for (DataIdx = 0; DataIdx < len; DataIdx++)
-	{
-	  *ptr++ = __io_getchar();
-	}
-
-   return len;
+    /* scanf calls _read() with len=1024, so eat one character at time */
+    *ptr = __io_getchar();
```
atigyi: would be good to figure out why scanf calls _read with 1024 for ethernet version and probably not? for wifi version. Or does wifi version scanf behave same?

examples/stm32/xively_demo_stm32f2xx_nucleo_eth/readme.md
atigyi: otherwise the comment is great to have!

examples/stm32/xively_demo_stm32f4xx_nucleo_eth/Src/demo_io.c
```diff
@@ -67,7 +67,8 @@ int8_t io_nucleoboard_init( void )
 int8_t io_read_button( void )
 {
     int32_t pin_state = BSP_PB_GetState( IO_NUCLEO_BUTTON_PIN );
-    ( pin_state == 1 ) ? ( pin_state = 0 ) : ( pin_state = 1 );
+    //( pin_state == 1 ) ? ( pin_state = 0 ) : ( pin_state = 1 ); /* The ethernet nucleo board's input is pulled up when pressed, not down like the wifi */
```
atigyi: ok, then delete this line

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Inc/provisioning.h
```diff
@@ -0,0 +1,8 @@
+#ifndef __PROVISIONING_H__
+#define __PROVISIONING_H__
+
+#include "user_data.h"
+
+int8_t provisioning_start( user_data_t* dst );
```
atigyi: `provisioning_start` makes me think to call `provisioning_finish` at the end when I finish the provisioning process. But actually this isn't true, it is not needed. I'd call this function `do_provisioning` or something better which reflects more its behaviour.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Inc/user_data.h
```diff
@@ -23,12 +29,15 @@ int8_t user_data_copy_from_flash( user_data_t* dst );
 int8_t user_data_save_to_flash( user_data_t* src );
 void user_data_printf( user_data_t* user_data );
 
+#if USER_DATA_WIFI_DEVICE
 #define user_data_set_wifi_encryption( data_ptr, encryption )         \
     ( data_ptr->wifi_client_encryption_mode = ( int32_t )encryption )
 #define user_data_set_wifi_ssid( data_ptr, ssid )                     \
     strncpy( data_ptr->wifi_client_ssid, ssid, USER_DATA_STR_SIZE )
 #define user_data_set_wifi_psk( data_ptr, psk )                       \
```
atigyi: here the `psk`-->`password`? as for eth device.

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -589,20 +596,50 @@ int main( void )
...
+        printf( "\r\n>> [ERROR] trying to copy user data from flash. Abort" );
         while ( 1 )
             ;
     }
+    printf( "\r\n>> User data retrieved from flash:" );
+    user_data_printf( &user_config );
+    fflush( stdout );
+
+    /* Provision the device if requested or if flash data is missing||corrupt */
+    int8_t provisioning_bootmode = io_read_button();
```
atigyi: add `const` for const-correctness if possible

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -589,20 +596,50 @@ int main( void )
...
+    if ( provisioning_bootmode || ( 0 > user_data_validate_checksum( &user_config ) ) )
+    {
+        ( provisioning_bootmode == 1 )
```
atigyi: here the `provisioning_bootmode` is used in two different ways as a boolean value. First is just testing the value the second checks its equality with 1. Two things here:
- 1st: select one type of testing, I'd prefer checking the equality to 1 since - as I know - it is up to the compiler? how it represents boolean TRUE and FALSE values. I it isn't necessary that 1 is TRUE and 0 is FALSE - although I haven't seen any other solution yet.
- 2nd: in 9thBit we agreed upon to put the tested value at front to eliminate the possibility of accidental assignment which is hardly debuggable, so I'd advise to use there `1 == provisioning_bootmode` at both places :)

examples/stm32/xively_demo_stm32f4xx_nucleo_wifi/Src/main.c
```diff
@@ -768,200 +804,6 @@ ex: printf("Wrong parameters value: file %s on line %d\r\n", file, line) */
 }
 #endif
 
-static inline void print_user_config_debug_banner( void )
```
atigyi: awesome you were able to remove these lines and put all into common code - even if it isn't consumed from common file by the two projects 👍

atigyi:
I've put some minor comments. As I see you already started the refactoring with this change, great. I will test this solution on a device too, then I will be back with feedback :) Great job!

examples/stm32/xively_demo_stm32f2xx_nucleo_eth/readme.md
Palantir555: Good point, will remove :+1:

examples/stm32/xively_demo_stm32f4xx_nucleo_eth/Inc/user_data.h
```diff
@@ -0,0 +1,48 @@
...
+int8_t user_data_copy_from_flash( user_data_t* dst );
+int8_t user_data_save_to_flash( user_data_t* src );
+void user_data_printf( user_data_t* user_data );
+
+#if USER_DATA_WIFI_DEVICE
+#define user_data_set_wifi_encryption( data_ptr, encryption )         \
+    ( data_ptr->wifi_client_encryption_mode = ( int32_t )encryption )
+#define user_data_set_wifi_ssid( data_ptr, ssid )                     \
+    strncpy( data_ptr->wifi_client_ssid, ssid, USER_DATA_STR_SIZE )
+#define user_data_set_wifi_psk( data_ptr, psk )                       \
```
Palantir555: True. I thought it was the SDK's nomenclature, but it doesn't seem to be used that often there. Will rename 👍

examples/stm32/xively_demo_stm32f4xx_nucleo_eth/SW4STM32/syscalls.c
Palantir555: I copied that comment and implementation from the wifi's `syscalls.c`. I'd also like to know why this implementation can ignore the `len` param, but it fixed all my issues, so I didn't investigate further.

examples/stm32/xively_demo_stm32f4xx_nucleo_eth/Src/main.c
Palantir555: Correct. `console.c` implements the serial I/O support for different compilers

examples/stm32/xively_demo_stm32f4xx_nucleo_eth/Src/provisioning.c
```diff
@@ -0,0 +1,172 @@
...
+            break;
+        default:
+            printf( "\r\n\t[ERROR] Getting AP credentials from user" );
+            return -1;
+    }
+    return 0;
+}
+
+#if USER_DATA_WIFI_DEVICE
+/**
```
Palantir555: Good point 👍  I'll keep it in mind for the implementation of the `Common` folder

Palantir555:
@atigyi I've implemented the changes you suggested, thanks for the review!

examples/stm32/xively_demo_stm32f4xx_nucleo_eth/SW4STM32/syscalls.c
atigyi: 👍

PR #125, opened 2017-06-26
SFT Client for Firmware Updates on TI CC3200

DellaBitta:
[ Description ]
A new reference client for Xively Secure File Transfer operations, and tied it to the CC3200 firmware update SDK.  This client should be used as en example of how to achieve over the air file transfers over the Xively MQTT service.

The device logs firmware update progress and errors to the Xively Device Logs service, so that firmware update status can be tracked remotely.

This client will be the engine of a new tutorial that will be published to the Xively Developer Center in a few more weeks.

[ Reviewer ]
@atigyi 

[ QA ]
Compiled the client with multiple OTA updates using the Xively Service.
Ran code formatter on the Xively-created source files.

Tested against SDKs:
* TI CC3200 SDK 1.1 with Service Pack: 1.0.1.6-2.7.0.0
* TI CC3200 SDK 1.2

[ Release Notes ]
Secure File Transfer: A new TI CC3200 client that allows firmware updates over MQTT via the Xively SFT service.


examples/cc3200/xively_firmware_updates/main.c
```diff
@@ -0,0 +1,1249 @@
...
+    Button_IF_Init( pushButtonInterruptHandlerSW2, pushButtonInterruptHandlerSW3 );
+
+    /* Make the buttons to react on both press and release events */
+    MAP_GPIOIntTypeSet( GPIOA1_BASE, GPIO_PIN_5, GPIO_BOTH_EDGES );
+    MAP_GPIOIntTypeSet( GPIOA2_BASE, GPIO_PIN_6, GPIO_BOTH_EDGES );
+
+    /* Disable the interrupt for now */
+    Button_IF_DisableInterrupt( SW2 | SW3 );
+
+    xi_state_t ret_state = xi_initialize( g_wifi_device_credentials.xivelyAccountId,
```
atigyi: after merging `development` branch into this branch the `xi_initialize` function will have one less argument. Please do the merge to make this branch "fresh".

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
```diff
@@ -0,0 +1,1336 @@
...
+              device_id );
+
+    /* Accept the latest firmware if we get here. */
+    if ( sl_extlib_FlcIsPendingCommit() )
+    {
+        printf( "\n********** Committing this revision:\"%s\" **********\n",
+                firmware_revision );
+        sl_extlib_FlcCommit( FLC_COMMITED );
+        xi_publish_device_log( in_context_handle, "Firmware Committed",
+                               "Boot loader will now boot to new image", sft_status_ok );
```
atigyi: To be more clear I am missing a "from now on" like statement from this expression.

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
```diff
@@ -0,0 +1,1336 @@
...
+    snprintf( xi_ctopic, sizeof( xi_ctopic ), "xi/ctrl/v1/%s/cln", device_id );
+
+    /* xi_logtopic is the MQTT topic we PUB to to add log messages. */
+    snprintf( xi_logtopic, sizeof( xi_logtopic ), "xi/blue/v1/%s/d/%s/_log", account_id,
+              device_id );
+
+    /* Accept the latest firmware if we get here. */
+    if ( sl_extlib_FlcIsPendingCommit() )
+    {
+        printf( "\n********** Committing this revision:\"%s\" **********\n",
```
atigyi: So this is the place where the decision is made by the new FW that I am OK. We talked about this on our meetup, but I would like to mention here too that "committing the new FW" must be preceded by a let's say "new firmware self check". And if this self check returns ok, then FW can be committed. This is just a memo for me, nothing to do with this example now.

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
DellaBitta: 👍

examples/cc3200/xively_firmware_updates/main.c
DellaBitta: Nice catch, thank you.

examples/cc3200/xively_firmware_updates/main.c
```diff
@@ -0,0 +1,1249 @@
...
+#define XIVELY_PWD_KEY "password"
+
+/* Used to detect if the user has configured their xively_config.txt file
+ * on the device before the execution of the application */
+#define XIVELY_DEFAULT_CONFIG_FIELD_VALUE "<PASTE_XIVELY_CREDENTIAL_HERE>"
+#define WIFI_DEFAULT_CONFIG_FIELD_VALUE "<ENTER_WIFI_CREDENTIAL_HERE>"
+
+/* String template for dynamically constructing the various Xively topics
+ * that are used for this application. */
+#define XIVELY_TOPIC_LEN ( 128 )
```
atigyi: is 128 enough? can't customer easily enter longer topic name? I mean there is no enough large number but isn't 256 more safe? just asking

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
```diff
@@ -0,0 +1,1336 @@
...
+char firmware_filename[] = "firmware";
+char firmware_revision[] = "1.0";
+
+file_download_ctx_t download_ctx;
+
+/* Buffers to hold the formatted topics.
+ * Topics are mangled by account and device ids.
+ */
+char xi_stopic[128];
+char xi_logtopic[128];
```
atigyi: dangerous static lenghts, but probably good for example purposes

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
```diff
@@ -0,0 +1,1336 @@
...
+            break;
+        case sft_status_ok:
+            severity_str = "notice";
+            break;
+        default:
+            severity_str = "error";
+            break;
+    }
+
+    char device_log_buffer[400];
```
atigyi: DANGER ☢️ DANGER 400, overflow

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
```diff
@@ -0,0 +1,1336 @@
...
+/*
+ * TYPEDEFS
+ */
+typedef unsigned char uchar;
+typedef unsigned int uint;
+
+typedef enum {
+    sft_status_ok = 0, /* 0 */
+
+    sft_status_rebooting, /* 1 */
```
atigyi: isn't it less error prone to use value assignments for every value than put "expected" values into comments?

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
DellaBitta: snprintf prevents overflow.

atigyi:
So the code reaches its purpose: sample implementation of the SFT + FWU for CC3200. 

Although it uses static sized arrays which may be overflown. The CODEC / protocol / logic are in spaghetti connection, only single file update (firmware) is supported with burnt in revision. The fingerprint is suck to SHA256 selection. So this is a very special usage and implementation of the Xively SFT. But there is a very big advantage of this solution: it really DOES the Firmware Update and it is available NOW.

So I think it is ok to have it in our repo, good to merge. This will fill in the hole until the Xively C Client provides FWU inside.

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
DellaBitta: These are static and not up to the customer to use for their own topic buffers. snprintf prevents overflow, and the svc , cln and _log channels fit well within this limitation.

DellaBitta: I should use the xi_toipc_len define though!

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
DellaBitta: I think the error prone ness there is that you can accidentally duplicate an assignment.  What do you think?

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
atigyi: ok, true

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
atigyi: you also can duplicate a comment with same value, I think this is a marginal question but still neither is good solution and during a debug session relying on a comment value which can be misleading is dangerous. If not dangerous then costs time. A value assignment... is it better?

DellaBitta: Yeah, I think it's less about reading the value and instead about using it as an index into a string table for the device logs.  If there's a typo error instead of an autoassigned value, then the error will be mislogged.  I can make them assignments and be careful about it.

examples/cc3200/xively_firmware_updates/main.c
DellaBitta: This isn't a flexible example, this is to show firmware updates on the CC3200.  The topics required for this example to work are within the 128 character limit.  

The other example is the Product Launcher stencil that people should build generic products on.  

In the case that their topic is longer than this, then snprintf will not crash, but subscription will fail, which will lead people to realize toward the problem.  The alternative is to calculate how strong a channel + topic name would be and then use malloc to allocate the topic of that length, but I find that this sort of code is overly complicated for an example. I'd much rather people focus on learning SFT and the Xively Client State Machine than to have a fool proof iron-clad topic name generator.

Though, as we have more and more overlapping code, perhaps we could create a Common Xively Client Toolkit that has that abstracted away in a tutorial SDK of some sort. ...?

atigyi: ok, this is ok for the example 👍

PR #129, opened 2017-07-05
Commit elf binaries for ST ethernet demo

Palantir555:
[ Description ]
We used to have `.bin` binaries for the ethernet project and `.elf` for the WiFi one.

Only .elf binaries can be flashed using Eclipse's debugger, so the tier 1 tutorial for the ethernet project would fail (unless the user compiled the project themselves and obtained the .elf).

This PR takes care of that issue:

- Remove `.bin` files from the ethernet project
- Commit `.elf` files for the ethernet project
- Commit new `.elf` files for the WiFi project -- This wasn't necessary, but since we're updating binaries we might as well make sure we're running the latest code

[ JIRA ]

[ Reviewer ]
atigyi

[ QA ]
- Ethernet + A1 sensor board runs OK -- Programmed to flash (target->Program chip...), instead using the 'Play' button in Eclipse
- Ethernet + A2 sensor board runs OK -- Programmed to flash (target->Program chip...), instead using the 'Play' button in Eclipse
- WiFi + A1 sensor board runs OK -- Programmed to flash (target->Program chip...), instead using the 'Play' button in Eclipse
- WiFi + A2 sensor board runs OK -- Programmed to flash (target->Program chip...), instead using the 'Play' button in Eclipse

I cannot program the devices using the 'play' button in Eclipse. I get an error "Could not determine the version of Valgrind", and the process seems to be aborted. Target->Program chip... works fine.

The ethernet board's SNTP implementation is not verbose enough on failure. I've had a couple runs loop forever when it fails. I don't have the time to improve it right now, but we should add it to the backlog.

[ Release Notes ]
STM32F4 Nucleo demos - Updated .elf files and removed extraneous .bin files.


atigyi:
From file change perspective this is OK. :+1: Do these .elf files work? Have you tested them?
I will test one of the A1s tomorrow since I will demo it on Friday.
If they are tested, then Approved :boom:

DellaBitta:
Only suggestion is to mention the SDK that they were compiled against in either the directory naming structure or in a README.md in the directory itself.

Palantir555:
@atigyi I've tested all 4 binaries and updated the PR description. Please check out the issues I've documented there

Palantir555:
@DellaBitta I've added 2 README.md files with the SDK versions we used

examples/stm32/xively_demo_stm32f4xx_nucleo_eth/README.md
```diff
@@ -0,0 +1,10 @@
+Context
+=======
+
+This example was developed and tested, and compiled into the provided `.elf`
```
atigyi: :+1:

atigyi:
Well done @Palantir555 :100: Approved!!

PR #130, opened 2017-07-06
Feature/remove redundant cbor sources

atigyi:
[ Description ]
Removal of redundant CBOR and SHA256 library sources around examples and libxively itself. There were two FWU examples using their own compilation of CBOR and SHA256 solution even if libxively already contained the code. So compilation of CBOR and SHA256 sources for both of these examples were removed SHA256 is used through a new BSP CRYPT interface. CBOR headers are included from src/import as any other 3rd party tools are included.

[ JIRA ]

[ Reviewer ]
@DellaBitta

[ QA ]
Redundant codes were removed, build settings updated, both examples compile and link. So they were rebuilt and manual tested. Both examples do file download, CC3200 does the FWU itself too.

[ Release Notes ]
n/a


examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
```diff
@@ -497,24 +501,26 @@ void xi_publish_file_info( xi_context_handle_t in_context_handle )
...
-                     cn_cbor_string_create( firmware_filename, &err ), &err );
+    cn_cbor* a        = cn_cbor_array_create( CBOR_CONTEXT_PARAM_COMMA & err );
+    cn_cbor* cb_file1 = cn_cbor_map_create( CBOR_CONTEXT_PARAM_COMMA & err );
+    cn_cbor_map_put( cb_file1, cn_cbor_string_create( "N" CBOR_CONTEXT_PARAM, &err ),
+                     cn_cbor_string_create( firmware_filename CBOR_CONTEXT_PARAM, &err ),
+                     &err );
 
-    cn_cbor_map_put( cb_file1, cn_cbor_string_create( "R", &err ),
-                     cn_cbor_string_create( "-1", &err ), &err );
+    cn_cbor_map_put( cb_file1, cn_cbor_string_create( "R" CBOR_CONTEXT_PARAM, &err ),
```
DellaBitta: Are these in the correct position?  It's after the string but before the comma?

examples/cc3200/xively_firmware_updates/main.c
```diff
@@ -627,10 +627,9 @@ void parseCredentialsFromConfigFile()
         broker_name = XIVELY_DEFAULT_BROKER;
     }
 
-    static uint16_t broker_port = XIVELY_DEFAULT_PORT;
     if ( 0 == parseKeyValue( config_file_context, XIVELY_PORT_KEY, 0, &broker_port_str ) )
     {
-        if ( 0 == ( broker_port = atoi( broker_port_str ) ) )
+        if ( 0 == ( atoi( broker_port_str ) ) )
```
DellaBitta: hah, now you're fixing TI's code. :)

examples/cc3200/xively_firmware_updates/main.c
```diff
@@ -983,7 +982,7 @@ void SimpleLinkNetAppEventHandler( SlNetAppEvent_t* pNetAppEvent )
 
             SET_STATUS_BIT( g_ulStatus, STATUS_BIT_IP_AQUIRED );
 
-            /* Ip Acquired Event Data *.
+            /* Ip Acquired Event Data */
```
DellaBitta: sorry about that. :(

examples/common/rules.mk
```diff
@@ -27,13 +27,14 @@ XI_EXAMPLE_BIN := $(XI_EXAMPLE_BINDIR)/$(XI_EXAMPLE_NAME)
...
 XI_CLIENT_INC_PATH += $(CURDIR)/../../include
+XI_CLIENT_INC_PATH += $(CURDIR)/../../include/bsp
 XI_CLIENT_LIB_PATH ?= $(CURDIR)/../../bin/$(XI_TARGET_PLATFORM)
 
 XI_CLIENT_ROOTCA_LIST := $(CURDIR)/../../res/trusted_RootCA_certs/xi_RootCA_list.pem
 
 XI_FLAGS_INCLUDE += $(foreach i,$(XI_CLIENT_INC_PATH),-I$i)
 
-XI_FLAGS_COMPILER ?= -Wall -Werror -Wno-pointer-arith -Wno-format -fstrict-aliasing -Os -Wextra
+XI_FLAGS_COMPILER ?= -Wall -Werror -Wno-pointer-arith -Wno-format -fstrict-aliasing -Os -Wextra -DUSE_CBOR_CONTEXT
```
DellaBitta: What does this flag do?

examples/firmware_update_example/src/xi_sft_client.c
```diff
@@ -135,7 +139,7 @@ void on_sft_message( xi_context_handle_t in_context_handle,
                                  params->message.temporary_payload_data_length );
 
             /* Figure out what the packet type is and then call the appropriate parser */
-            cb = cn_cbor_decode( payload_data, payload_length, 0 );
+            cb = cn_cbor_decode( payload_data, payload_length CBOR_CONTEXT_PARAM, NULL );
```
DellaBitta: Could you briefly leave a reply to this message as to what the difference is between this CBOR_CONTEXT_PARAM and the COMMA version?

make/mt-config/mt-config.mk
```diff
@@ -191,8 +191,8 @@ XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/datastructures
 XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/mqtt/codec
 XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/mqtt/logic
 XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/control_topic
+XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/../import/cn-cbor/src
```
DellaBitta: use LIBXIVELY_SRC

make/mt-config/mt-config.mk
```diff
@@ -221,7 +221,7 @@ XI_INCLUDE_FLAGS += $(foreach platformdep,$(XI_PLATFORM_MODULES) \
 XI_INCLUDE_FLAGS += $(foreach d, $(LIBXIVELY_INTERFACE_INCLUDE_DIRS), -I$d)
 
 XI_INCLUDE_FLAGS += -I$(LIBXIVELY_SOURCE_DIR)
-XI_INCLUDE_FLAGS += -I$(LIBXIVELY_SOURCE_DIR)/cbor/cn-cbor/include
+XI_INCLUDE_FLAGS += -I$(LIBXIVELY_SOURCE_DIR)/../import/cn-cbor/include
```
DellaBitta: and here.

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
atigyi: yes, and yes, it isn't an ordinary solution from the CN-CBOR C codec, but works and has to be used like this. The macro contains the coma. And the macro can be empty as well when no context is used. That is why it looks like this.

examples/cc3200/xively_firmware_updates/main.c
atigyi: it was an unused variable warning

examples/cc3200/xively_firmware_updates/main.c
atigyi: I don't think it is a fault. It is more like a great experience how things can go wrong in code. So this remained a valid (compiler accepted this) C code. Miracle.

examples/common/rules.mk
atigyi: this flag turns on the definitions for the CBOR context macros: CBOR_CONTEXT_PARAM.
And since the example uses CBOR from libxively itself it has to be aligned with the libxively CN-CBOR flags.

examples/firmware_update_example/src/xi_sft_client.c
atigyi: you can easily check it at the place of the definition which quickly explains it:
#define CBOR_CONTEXT_PARAM , context
#define CBOR_CONTEXT_PARAM_COMMA context ,
they differ in the position of the comma: before or after

make/mt-config/mt-config.mk
atigyi: :+1: good catch, thanks

make/mt-config/mt-config.mk
atigyi: :+1:

examples/common/rules.mk
atigyi: ah, yes, and the CBOR CONTEXT is used to pass the memory management function. This makes CBOR using platform dependent allocators / deallocators.

DellaBitta: OK, so this preoprocessor doesn't actually alter any function signatures.  The scary thing is that we're blindly adding it regardless of library configuration due to it currently being benign -- it only adds some preprocessor definitions but doesn't change signatures.  If it ever did change signatures then including it here but not including it in the library when built will cause unexpected crashes.  

That's why I ask.

examples/firmware_update_example/src/xi_sft_client.c
DellaBitta: well that's confusing, but ok!

examples/common/rules.mk
atigyi: the preproc def does alter function headers, it adds an extra context parameter.

atigyi: but anyway what you say is true. One can do it wrong. With building libxively and the example / application with different CBOR flags. But this is already true for wolfSSL's lib.

examples/firmware_update_example/src/xi_sft_client.c
atigyi: agree, not a usual solution but this is how CN-CBOR works. We can move to other CBOR lib but I don't know whether it supports customizable memory handling.

PR #134, opened 2017-07-27
Secure File Transfer: FILE BSP

atigyi:
[ Description ]
Writing out data received through SFT from Xively Broker. Writing the data is done through a new BSP FILE API which has now two implementations: POSIX and CC3200. This FILE BSP is responsible to handle FW binary and ordinary binaries as well. Another new BSP FWU helps to decide whether a file is firmware.

[ JIRA ]
[ SFT ] implement FILE BSP
https://jira.3amlabs.net/browse/XCL-2998

[ Reviewer ]
@DellaBitta

[ QA ]
Manual test is done for both implementations POSIX and CC3200. Both download a single update package CC3200 reboots and starts the new FW just received. Obviously macOS does not reboot, it's FWU BSP is implemented with "printf" functions. Another test with 3 files in the update package was done. Two "config" files and one FW: firmware.bin file. CC3200 handled this too.

[ Release Notes ]
n/a


src/libxively/control_topic/xi_control_message.h
```diff
@@ -28,9 +28,12 @@ typedef enum xi_control_message__sft_file_status_phase_e {
...
 typedef enum xi_control_message__sft_file_status_code_e {
-    XI_CONTROL_MESSAGE__SFT_FILE_STATUS_CODE_ERROR__UNEXPECTED_FILE_CHUNK_ARRIVED = -100,
-    XI_CONTROL_MESSAGE__SFT_FILE_STATUS_CODE_SUCCESS                              = 0,
-    XI_CONTROL_MESSAGE__SFT_FILE_STATUS_CODE_ADD_ERRORS_HERE                      = 1,
+    XI_CONTROL_MESSAGE__SFT_FILE_STATUS_CODE_ERROR__FILE_CLOSE            = -4,
+    XI_CONTROL_MESSAGE__SFT_FILE_STATUS_CODE_ERROR__FILE_WRITE            = -3,
+    XI_CONTROL_MESSAGE__SFT_FILE_STATUS_CODE_ERROR__FILE_OPEN             = -2,
+    XI_CONTROL_MESSAGE__SFT_FILE_STATUS_CODE_ERROR__UNEXPECTED_FILE_CHUNK = -1,
+    XI_CONTROL_MESSAGE__SFT_FILE_STATUS_CODE_SUCCESS                      = 0,
+    XI_CONTROL_MESSAGE__SFT_FILE_STATUS_CODE_ADD_ERRORS_HERE              = 1,
```
DellaBitta: shouldn't this be -5?

src/libxively/io/fs/posix/xi_fs_posix.c
```diff
@@ -1,97 +1,11 @@
-/* Copyright (c) 2003-2016, LogMeIn, Inc. All rights reserved.
+/* Copyright (c) 2003-2017, LogMeIn, Inc. All rights reserved.
  *
```
DellaBitta: wow, looks like this file's new years resolution was to lose weight.

src/libxively/io/fs/xi_fs_api.h
```diff
@@ -10,47 +10,8 @@
 #include <stdlib.h>
```
DellaBitta: Now that you have me thinking of copyright dates, please update this file's copyright date.

src/bsp/platform/cc3200/xi_bsp_fwu_cc3200.c
```diff
@@ -0,0 +1,56 @@
...
+
+#include <simplelink.h>
+#include <flc_api.h>
+
+#include "hw_types.h"
+#include "rom_map.h"
+#include "prcm.h"
+#include "utils.h"
+
+uint8_t xi_bsp_fwu_is_firmware( const char* const resource_name )
```
atigyi: xi_bsp_fwu_ is the prefix for all functions here

atigyi: David.team_question_is_this_acceptable_function_name( "xi_bsp_fwu_is_this_firmware" );

src/bsp/platform/cc3200/xi_bsp_io_fs_cc3200.c
```diff
@@ -0,0 +1,213 @@
...
+#include <xi_bsp_fwu.h>
+#include <string.h>
+#include <stdio.h>
+
+#include <simplelink.h>
+#include <gpio_if.h>
+
+static _i32 firmware_file_handle_last_opened = 0;
+
+#define XI_DEBUG__FOR_FIRMWARE_UPDATE_TESTING_PURPOSES
```
atigyi: ok, although I don't agree it belongs to header, it is a .c local define which should have effect only on this compilation unit. Headers may be included in other .c / compilation units. I'll move it up.

src/bsp/platform/cc3200/xi_bsp_io_fs_cc3200.c
```diff
@@ -0,0 +1,213 @@
...
+        {
+            status = 0;
+        }
+        sl_FsClose( lFileHandle, 0, 0, 0 );
+    }
+
+    return status;
+}
+
+#endif
```
atigyi: 👍

src/bsp/platform/cc3200/xi_bsp_io_fs_cc3200.c
```diff
@@ -0,0 +1,213 @@
...
+    return XI_STATE_OK;
+}
+
+#define XI_BSP_IO_FS_READ_BUFFER_SIZE 1024
+
+xi_state_t xi_bsp_io_fs_read( const xi_fs_resource_handle_t resource_handle,
+                              const size_t offset,
+                              const uint8_t** buffer,
+                              size_t* const buffer_size )
+{
```
atigyi: very good question! It is NOT used yet. It'll be used in dynamic revision when the device will create a checksum of the filecontent. But right, it isn't used yet, and it is not recommended to have this implementation in this PR, but it is here, hope it works, I'll test it in next story (revision handling).

src/bsp/platform/cc3200/xi_bsp_io_fs_cc3200.c
DellaBitta: true, true!

src/bsp/platform/cc3200/xi_bsp_io_fs_cc3200.c
```diff
@@ -0,0 +1,213 @@
...
+                               const size_t buffer_size,
+                               const size_t offset,
+                               size_t* const bytes_written )
+{
+    { /* led indicator of file writes */
+#ifdef XI_DEBUG__FOR_FIRMWARE_UPDATE_TESTING_PURPOSES
+        if ( IMG_ACT_USER1 == sBootInfo.ucActiveImg )
+        {
+            GPIO_IF_LedOff( MCU_ORANGE_LED_GPIO );
+            GPIO_IF_LedToggle( MCU_GREEN_LED_GPIO );
```
atigyi: This code / feedback does not touch RED LED at all. This blinks either the green or the orange during file writes depending on the current FW's file position (1 or 2).

src/bsp/platform/cc3200/xi_bsp_io_fs_cc3200.c
```diff
@@ -0,0 +1,213 @@
...
+    {
+        return ( 0 == sl_FsClose( resource_handle, NULL, NULL, 0 ) ) ? XI_STATE_OK
+                                                                     : XI_FS_CLOSE_ERROR;
+    }
+}
+
+xi_state_t xi_bsp_io_fs_remove( const char* const resource_name )
+{
+    return ( 0 == sl_FsDel( ( const _u8* )resource_name, NULL ) ) ? XI_STATE_OK
+                                                                  : XI_FS_ERROR;
```
atigyi: hm, I don't know.

atigyi: Should some self kill logic reside here? Anyway how to check it? The only way to check it is to call the xi_bsp_fwu_is_this_firmware function. But it tells "yes" only for "firmware.bin" string. And actually the firmware on CC3200 is /sys/mcuimg1.bin or /sys/mcuimg2.bin. And yes, I could test against these strings but maybe the inactive firmware is asked to be removed which might be ok. The above copied function - which checks active FW is private inside the SDK - is just a copy, new SDK could update its body.
So two things: 
1 - I don't know whether it is the responsibility of a FILE BSP to prevent application to be dumb.
2 - Not that easy to deny only active firmware deletion.

So what to do? If you still advice to implement it I can create a heuristic which prevents deletion in 98% of the cases :) Although I think it is an unnecessary thing especially because it is more complex than its value. Especially at this stage of the FWU feature.

src/bsp/platform/posix/xi_bsp_io_fs_posix.c
```diff
@@ -0,0 +1,328 @@
...
+ * @param fp
+ * @return 1 if list element is the one with the matching fp 0 otherwise
+ */
+static uint8_t
+xi_fs_posix_file_list_cnd( xi_fs_posix_file_handle_container_t* list, FILE* fp )
+{
+    assert( NULL != list );
+    assert( NULL != fp );
+
+    return ( list->posix_fp == fp ) ? 1 : 0;
```
atigyi: This function is a functor - naming is from C++. Used in lines 200, 256 and 290 by XI_LIST_FIND macros. The FIND macros uses this function to check each element whether it is the searched one. Actually this is like a "bool std::operator==( const A& op1, const A& op2);" function in C++.
And yes, I think the list parameter should be renamed to list_element, I'll do this.

src/bsp/platform/posix/xi_bsp_io_fs_posix.c
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    xi_state_t ret = XI_STATE_OK;
+
+    struct stat stat_struct;
+    memset( &stat_struct, 0, sizeof( stat_struct ) );
+
+    int res = stat( resource_name, &stat_struct );
+
+    /* verification of the os function result */
```
atigyi: I wrote it here, but I don't think we should mention it everywhere it is used. Otherwise we could describe each function at its call point and not at the declaration.

src/bsp/platform/posix/xi_bsp_io_fs_posix.c
```diff
@@ -0,0 +1,328 @@
...
+    memset( new_entry, 0, sizeof( xi_fs_posix_file_handle_container_t ) );
+
+    /* store the posix file pointer */
+    new_entry->posix_fp = fp;
+
+    /* add the entry to the database */
+    XI_LIST_PUSH_BACK( xi_fs_posix_file_handle_container_t, xi_fs_posix_files_container,
+                       new_entry );
+
+    /* make sure that size does not matter */
```
atigyi: 👍

src/libxively/control_topic/xi_control_message.h
atigyi: ah, indeed 👍

src/libxively/io/fs/posix/xi_fs_posix.c
atigyi: :) the weight was picked up by xi_bsp_io_fs_posix.c.

src/libxively/io/fs/xi_fs_api.h
atigyi: ok, done, but we should do it for every file, or we wait 5 months and 2017 will be gone.

DellaBitta: haha I like the latter, you?  :)  I think so long as we update the files that we edit, then the coypright still holds for all files.

src/bsp/platform/cc3200/xi_bsp_io_fs_cc3200.c
DellaBitta: I don't know.  I think just adding the is firmware xi_bsp_fwu_is_this_firmware  check makes sense.  In this board's case I think the deletion will fail anyway, but I feel the check has more of a safety-first story to the reader, in case anyone uses this as a basis for another implementation.

src/bsp/platform/cc3200/xi_bsp_io_fs_cc3200.c
DellaBitta: Gotcha!

src/bsp/platform/cc3200/xi_bsp_io_fs_cc3200.c
DellaBitta: Ok, just making sure I wasn't missing anything. :)

DellaBitta:
👍

src/bsp/platform/cc3200/xi_bsp_io_fs_cc3200.c
atigyi: recommended ---> required, sorry

src/bsp/platform/cc3200/xi_bsp_io_fs_cc3200.c
atigyi: the xi_bsp_fwu_is_this_firmware function tests agains "firmware.bin" but the firmware is actually "/sys/mcuimg1.bin" or "/sys/mcuimg2.bin" or without number in case of no FW update. 
So I still don't know how to implement this.

src/libxively/io/fs/xi_fs_api.h
atigyi: 👍

atigyi:
thx, let's discuss the xi_bsp_io_file_remove's CC3200 implementation in Slack.

PR #136, opened 2017-08-02
Feature/sft revision handling

atigyi:
[ Description ]
Non-volatile storage of file revisions in a SFT update package. FILE BSP is used to store revision to each individual update file in a new file suffixed with .xirev. Since POSIX and CC3200 FILE BSPs are ready this revision non-volatile storage works for both of these platforms. Hopefully it will for all platforms which can implement the FILE BSP module.

[ JIRA ]
[ SFT ] file revision handling
https://jira.3amlabs.net/browse/XCL-3000

[ Reviewer ]
@DellaBitta

[ QA ]
Manual tests were done on POSIX and CC3200 platforms. *.xirev files were written out correctly and after reset they were read correctly since their content blocked update on Xively Broker side (probably the revision matched thus broker did not send back FILE_UPDATE_AVAILABLE).
Auto integration tests will come soon.

[ Release Notes ]
n/a


src/tests/itests/xi_itest_sft.c
```diff
@@ -153,7 +153,15 @@ void xi_itest_sft__remove_files( const char** filenames, uint16_t files_count )
     uint16_t id_file = 0;
     for ( ; id_file < files_count; ++id_file )
     {
+        /* delete files generated during tests (files received from mock broker) */
         xi_bsp_io_fs_remove( filenames[id_file] );
+
+#if 1
+        /* delete files containing revisionstrings */
```
DellaBitta: remove #if ?

src/libxively/sft/xi_sft_logic.c
```diff
@@ -268,6 +270,9 @@ xi_sft_on_message( xi_sft_context_t* context, xi_control_message_t* sft_message_
                             XI_CONTROL_MESSAGE__SFT_FILE_STATUS_CODE_SUCCESS );
                     }
 
+                    xi_sft_revision_set( context->update_current_file->name,
```
DellaBitta: I assume checksum would be before this eventually?

DellaBitta:
Looks ok.  Looks good actually.  Some small questions for my own enjoyment, and maybe the removal of an #if #endif

src/tests/itests/xi_itest_sft.c
atigyi: unfortunately you are right. I've left it there intentionally since it is good to see tests using the previous run's saved revision. But yeah, I'll create tests to cover checking the .xirev files. So yes, I'll remove this, thx.

src/libxively/sft/xi_sft_logic.c
atigyi: hm, good question, haven't thought about this yet. And maybe checksum should happen even before writing the file itself with final filename onto the flash storage. Since a broken content may overwrite the previous old but intact file.

PR #142, opened 2017-09-08
Feature/xi bsp io fs enum rename

DellaBitta:
[ Description ]
XI_BSP_IO_FS Enumeration and Struct Type renames.

Structs and Enums now match the NET and TLS BSP nomenclature.

[ JIRA ]
XCL-3039 rename enum values in xi_bsp_io_fs.h
https://jira.3amlabs.net/browse/XCL-3039

[ Reviewer ]
@atigyi 

[ QA ]
Travis and Local Builds

[ Release Notes ]
n/a

### DIFF
--- a/include/bsp/xi_bsp_io_fs.h
+++ b/include/bsp/xi_bsp_io_fs.h
@@ -27,7 +27,6 @@
 #include <stdint.h>
 #include <xively_error.h>
 
-
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -69,9 +68,6 @@ typedef intptr_t xi_bsp_io_fs_resource_handle_t;
     /** error reported when file cannot be closed */
     XI_BSP_IO_FS_CLOSE_ERROR = 11,
 } xi_bsp_io_fs_state_t;
-
-/* helper function that translates the errors to the xi_bsp_io_fs_state_t values */
-xi_state_t xi_fs_bsp_io_fs_2_xi_state( xi_bsp_io_fs_state_t bsp_state_value );
 
 /**
  * @enum xi_bsp_io_fs_resource_type_t
--- a/include/bsp/xi_bsp_io_fs.h
+++ b/include/bsp/xi_bsp_io_fs.h
@@ -32,35 +32,35 @@
 extern "C" {
 #endif
 
-typedef intptr_t xi_fs_resource_handle_t;
+typedef intptr_t xi_bsp_io_fs_resource_handle_t;
 
-#define XI_FS_INVALID_RESOURCE_HANDLE -1
-#define xi_fs_init_resource_handle() XI_FS_INVALID_RESOURCE_HANDLE
+#define XI_BSP_IO_FS_INVALID_RESOURCE_HANDLE -1
+#define xi_bsp_io_fs_init_resource_handle() XI_BSP_IO_FS_INVALID_RESOURCE_HANDLE
 
 /**
- * @enum xi_fs_resource_type_t
+ * @enum xi_bsp_io_fs_resource_type_t
  * @brief describes types of resources that are availible through this API.
  * These types were created in order to differenciate types based on their
  * security class.
  */
-typedef enum {
-    XI_FS_CERTIFICATE = 0, /**< 0 **/
-    XI_FS_CREDENTIALS,     /**< 1 **/
-    XI_FS_CONFIG_DATA      /**< 2 **/
-} xi_fs_resource_type_t;
+typedef enum xi_bsp_io_fs_resource_type_e {
+    XI_BSP_IO_FS_CERTIFICATE = 0, /**< 0 **/
+    XI_BSP_IO_FS_CREDENTIALS,     /**< 1 **/
+    XI_BSP_IO_FS_CONFIG_DATA      /**< 2 **/
+} xi_bsp_io_fs_resource_type_t;
 
 /*
- * @name xi_fs_stat_s
+ * @name xi_bsp_io_fs_stat_s
  * @brief Information that the Xively Client needs returned
  * when xi_bsp_io_fs_stat() is called.
  */
-typedef struct xi_fs_stat_s
+typedef struct xi_bsp_io_fs_stat_s
 {
     size_t resource_size;
-} xi_fs_stat_t;
+} xi_bsp_io_fs_stat_t;
 
 /**
- * @enum xi_fs_open_flags_t
+ * @enum xi_bsp_io_fs_open_flags_t
  * @brief Abstracted values that represent the various types of file operations
  * that should be passed to the underlying system when opening a file resource
  * via xi_bsp_io_fs_open().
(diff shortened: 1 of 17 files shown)


include/bsp/xi_bsp_io_fs.h
```diff
@@ -32,48 +32,131 @@
...
+} xi_bsp_io_fs_state_t;
+
+/* helper function that translates the errors to the xi_bsp_io_fs_state_t values */
+/* TODO: remove the need to have this inline, as the function has a substantial jump
+ * table. */
+static inline xi_state_t xi_fs_bsp_io_fs_2_xi_state( xi_bsp_io_fs_state_t bsp_state_value )
+{
+    xi_state_t ret = XI_STATE_OK;
+
+    switch ( bsp_state_value )
```
atigyi: what is the purpose of the new, separated BSP enum introduction? I thought the purpose is to hide the xively_error.h values. But they are not hidden. Other thought was to prevent the customer implemented BSP IO FS layer to return non IO FS related errors e.g. XI_SOCKET_GETHOSTBYNAME_ERROR or something total non-related. But as I see the BSP is responsible to convert IO_FS_ERRORS to libxively_ERROS.
I would do one of the following:
1. let BSP IO FS use xively_error.h errors
2. move the conversion (implementation of this function) to xi_fs_posix.c

src/bsp/platform/posix/xi_bsp_io_fs_posix.c
```diff
@@ -26,24 +26,25 @@ const size_t xi_fs_buffer_size = 1024;
...
 
 /* local list of open files */
-static xi_fs_posix_file_handle_container_t* xi_fs_posix_files_container;
+static xi_bsp_io_fs_posix_file_handle_container_t* 
+    xi_bsp_io_fs_posix_files_container;
 
-/* helper function that translates the errno errors to the xi_state_t values */
-static xi_state_t xi_fs_posix_errno_2_xi_state( int errno_value )
+/* translates bsp errno errors to the xi_bsp_io_fs_state_t values */
+xi_bsp_io_fs_state_t xi_bsp_io_fs_posix_errno_2_xi_state( int errno_value )
```
atigyi: xi_fs_posix_errno_2_xi_bsp_io_fs_error instead?

src/bsp/platform/posix/xi_bsp_io_fs_posix.c
```diff
@@ -52,39 +53,39 @@ static xi_state_t xi_fs_posix_errno_2_xi_state( int errno_value )
...
+            ret = XI_BSP_IO_FS_ERROR;
     }
 
     return ret;
 }
 
 /* helper function that translates posix stat to xi stat */
-static xi_state_t xi_fs_posix_stat_2_xi_stat( const struct stat* const posix_stat,
-                                              xi_fs_stat_t* const xi_stat )
+static xi_bsp_io_fs_state_t xi_bsp_io_fs_posix_stat_2_xi_stat( const struct stat* const posix_stat,
```
atigyi: xi_fs_posix_stat_2_xi_bsp_io_fs_posix_stat?

atigyi: + ohnos there are two types with slightly same name, easy to mix them up:
- xi_bsp_io_fs_state_t
- xi_bsp_io_fs_stat_t
No problem here, just hmm... I don't know. It can stay if you are ok with them.

src/libxively/io/fs/posix/xi_fs_posix.c
```diff
@@ -15,7 +18,19 @@ xi_state_t xi_fs_stat( const void* context,
...
     XI_UNUSED( resource_type );
 
-    return xi_bsp_io_fs_stat( resource_name, resource_stat );
+    if ( NULL == resource_stat || NULL == resource_name )
+    {
+        return XI_INVALID_PARAMETER;
+    }
+    
+    xi_bsp_io_fs_stat_t bsp_io_fs_resource_stat;
+    xi_state_t result = xi_fs_bsp_io_fs_2_xi_state( xi_bsp_io_fs_stat( resource_name, &bsp_io_fs_resource_stat ) );
```
atigyi: yeah, so the implementation of this converter function should be in this file. At least I think so.

src/libxively/sft/xi_sft_logic_file_chunk_handlers.c
```diff
@@ -40,10 +40,10 @@ xi_sft_on_message_file_chunk_process_file_chunk( xi_sft_context_t* context,
     /* write bytes through FILE BSP */
     size_t bytes_written = 0;
 
-    state =
+    state = xi_fs_bsp_io_fs_2_xi_state(
```
atigyi: So, again, why do we do this conversion? Why isn't BSP IO FS allowed to return xi_state_t? The xively_error.h is already accessible, it is on the API? Doesn't this sound like an unnecessary overhead?

include/bsp/xi_bsp_io_fs.h
DellaBitta: Yeah, I see your point but I did it this way for a few of reasons:


1. I used the Networking BSP and TLS BSP as a standard method of practice, and each of them its own BSP Error codes.  I was trying to be consistent.

2. It creates a very succinct vocabulary for filling out a particular BSP.  As an engineer I know exactly what’s expected due to the limited number of enumerations.  This is important for easing the development of new systems.  You and I know might know what sort of error codes the invoking code expects back, what are good errors to return, etc, but are new engineers coming to the code base so well acquainted with what to do?

2a. If we use the main error codes, which is an extensive list, it gets confusing for developers.  I’ve been in this scenario before where there were global error enumerations, butI don’t know which were valid errors to return for a given subsystem and I didn’t know if I could cause problems returning wrong error codes.  I also took a lot of time guessing what I SHOULD be returning, etc, where it woul be much nicer to have that list be a small and focused as possible.

Mostly for 1… but I do see a lot of value in 2…. Too.

src/bsp/platform/posix/xi_bsp_io_fs_posix.c
DellaBitta: ok, sure!

src/libxively/io/fs/posix/xi_fs_posix.c
DellaBitta: Yep! Thanks!  👍

src/libxively/sft/xi_sft_logic_file_chunk_handlers.c
DellaBitta: See my main reply comment above.

src/libxively/io/fs/posix/xi_fs_posix.c
```diff
@@ -6,6 +6,58 @@
 
 #include "xi_fs_header.h"
 #include "xi_macros.h"
+#include "xi_bsp_io_fs.h"
+
+const size_t xi_fs_buffer_size = 1024;
+
+xi_state_t xi_fs_bsp_io_fs_2_xi_state( xi_bsp_io_fs_state_t bsp_state_value )
```
atigyi: thx for moving this function! Could you please put this function into a file avoiding the duplication in files xi_fs_memory.c and xi_fs_posix.c? I know it is planned to be removed but still who knows how long it'll exist? It may go out of sync for mem and file implementation if two versions exist. Thanks.

atigyi:
Approved 🥇

PR #143, opened 2017-09-11
SFT feature minor changes during testing

atigyi:
[ Description ]
SFT minor features:
- introduction of a new SFT feature build switch, turning it on/off activates/deactivates the compilation of SFT related source files
- minor fix in the attributes of the checksum generator function

[ JIRA ]
[ SFT ] small but important things left behind
https://jira.3amlabs.net/browse/XCL-3003

[ Reviewer ]
@DellaBitta

[ QA ]
Auto tests still run and pass. POSIX FWU is tested.

[ Release Notes ]
n/a


make/mt-config/mt-config.mk
```diff
@@ -222,6 +223,16 @@ ifneq (,$(findstring control_topic,$(CONFIG)))
...
+ifneq (,$(findstring secure_file_transfer,$(CONFIG)))
+	XI_INCLUDE_FLAGS += -I$(LIBXIVELY_SRC)/import/cn-cbor/include
+
+	XI_SRCDIRS += $(LIBXIVELY_SRC)/import/cn-cbor/src
+	XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/cbor
+	XI_SRCDIRS += $(LIBXIVELY_SOURCE_DIR)/sft
+
+	XI_SECURE_FILE_TRANSFER_ENABLED := 1
+endif
+
```
DellaBitta: :+1:

src/bsp/platform/cc3200/xi_bsp_fwu_cc3200.c
```diff
@@ -29,6 +46,26 @@ xi_state_t xi_bsp_fwu_on_new_firmware_ok()
...
+{
+    if ( sl_extlib_FlcIsPendingCommit() )
+    {
+        sl_extlib_FlcCommit( FLC_NOT_COMMITED );
+    }
+
+    _reboot_device();
+
+    /* Control should never reach this */
+    return XI_STATE_OK;
```
DellaBitta: Shouldn't this be an error?

src/libxively/sft/xi_sft_revision.c
```diff
@@ -163,15 +163,27 @@ xi_state_t xi_sft_revision_firmware_ok()
     char* xi_firmware_resource_name = firmware_update_data;
     char* xi_firmware_revision      = firmware_update_data;
 
-    strtok_r( xi_firmware_revision, "\n", &xi_firmware_revision );
+    while ( '\n' != *xi_firmware_revision && 0 != *xi_firmware_revision )
+    {
+        ++xi_firmware_revision;
+    }
+
```
DellaBitta: Just making sure that you know that this string is always null terminated?  It makes me nervous that we don't put a cap on this increment loop based on the size of the string.

DellaBitta: Perhaps add an out parameter to xi_sft_revision_read_string_from_file that is the buffer size or the string length, so that you can do a limit on the number of bytes compared here.

atigyi: A string is string because a zero indicates its end. Otherwise it is a series of bytes. The function name xi_sft_revision_read_string_from_file says "...read_string..." so the output is a string which is null terminated by definition. Other solution is to rename the function and call it xi_sft_revision_read_data_from_file or ...read_bytes... this function must have a length output parameter.
So either of these is sufficient. I don't think two indicators should carry the content end information. Here I chose the terminating zero since I exactly know that the file content does not have zero since it is a string, the zero is the end.
So I wouldn't touch this only if you preferred the "...read_bytes..." function renaming. But this wouldn't make sense because this file contains a string, its purpose is to contain a string.

src/bsp/platform/cc3200/xi_bsp_fwu_cc3200.c
atigyi: 👍 changed

src/libxively/sft/xi_sft_revision.c
DellaBitta: I bring this up because I know that when people code for embedded devices, they put in safeguards to ensure they don't do buffer overruns.  things like strncmp and strncpy and strncat exist for these reasons.  

But now after looking closer I see that the buffer guarantees a null terminator in there.  So I'm ok with it.

Issue #153, opened 2017-10-06
Fails fo Building the MQTT and TLS libraries  under Windows 


WilliamFrasson:
Hi All,
I got these messages try to complie  wolfssl  under windows 10 using msys shell:
"error: 'WOLFSSL_CERT_MANAGER {aka struct WOLFSSL_CERT_MANAGER}' has no member named 'ocsp_stapling'"

here below the log
what i miss or did wrong?
thanks in advaced

--------------
```
$ make PRESET=ESP32

'# Using build PRESET: "ESP32" to set CONFIG and TARGET variables:
.    XI_BSP_PLATFORM: [esp32]
.    XI_BSP_TLS:      [wolfssl]

'# Xively C Client

. run 'make help' for more information, including a list of valid Build PRESETS.
.    CONFIG:          [memory_fs-tls_bsp]
.    TARGET:          [esp32-static-release]
.    COMPILER:        [xtensa-esp32-elf-gcc]

make[1]: ingresso nella directory "/c/esp32/xively-client-c/examples/esp32/xively_demo/wolfssl-make"
  [CC] /c/esp32/xively-client-c/src/import/tls/wolfssl/src/ssl.c
C:/esp32/xively-client-c/src/import/tls/wolfssl/src/ssl.c: In function 'wolfSSL_CertManagerFree':
C:/esp32/xively-client-c/src/import/tls/wolfssl/src/ssl.c:2367:19: error: 'WOLFSSL_CERT_MANAGER {aka struct WOLFSSL_CERT_MANAGER}' has no member named 'ocsp_stapling'
             if (cm->ocsp_stapling)
                   ^
C:/esp32/xively-client-c/src/import/tls/wolfssl/src/ssl.c:2368:28: error: 'WOLFSSL_CERT_MANAGER {aka struct WOLFSSL_CERT_MANAGER}' has no member named 'ocsp_stapling'
                 FreeOCSP(cm->ocsp_stapling, 1);
                            ^
C:/esp32/xively-client-c/src/import/tls/wolfssl/src/ssl.c: In function 'wolfSSL_CertManagerEnableOCSPStapling':
C:/esp32/xively-client-c/src/import/tls/wolfssl/src/ssl.c:4793:15: error: 'WOLFSSL_CERT_MANAGER {aka struct WOLFSSL_CERT_MANAGER}' has no member named 'ocsp_stapling'
         if (cm->ocsp_stapling == NULL) {
               ^
[...]

PR #155, opened 2017-10-09
Fix WolfSSL build for ESP32 platform

Palantir555:
[ Description ]
WolfSSL builds fine with the existing settings in macOS, Ubuntu and Windows 10, so I'm not sure why it's not working for other users.

@atigyi 's suggested solution of removing the NO_WOLFSSL_SERVER fixed the issue for other users, and I've validated it still works on my machine, so let's just push the fix and move to the next task.

[ JIRA ]

[ Reviewer ]

[ QA ]
Built the binaries on macOS, WilliamFrasson was able to solve his issue #153 with this fix.

[ Release Notes ]
Fix WolfSSL build for the ESP32 platform.


atigyi:
Approved, 👍

PR #158, opened 2017-10-12
remove early phase SFT posix example, this already mislead a prospect :(

atigyi:
[ Description ]
This change removes an early phase FWU / SFT example implementation. Since the integrated FWU / SFT is released there is no need for PoC examples to exists especially since it is misleading for visitors.

[ Reviewer ]

[ QA ]
Build all other examples from examples dir, which means the build all does not fail because of the removal. This example wasn't even in the list of the all examples.

[ Release Notes ]
n/a


examples/firmware_update_example/Makefile
```diff
@@ -1,16 +0,0 @@
-# Copyright (c) 2003-2016, LogMeIn, Inc. All rights reserved.
-#
```
DellaBitta: 2017. 

Just kidding!

atigyi: :D oh man, this was good 🤣

PR #160, opened 2017-10-25
Feature/ESP32 Over The Air firmware updates

Palantir555:
[ Description ]

Fix SFT use-after-free bug. A file was being closed before its read buffer was copied, which caused occasional buffer corruptions.

Working OTA firmware updates for the ESP32:

1. Implemented the ESP32's file and fwu BSPs
2. Implement sdkconfig.defaults with default project configuration
3. Create a partition for a FAT Filesystem, to be consumed by the SFT logic
4. Updated the example to use OTA firmware updates
5. Updated ESP32 preset to enable the control topic and SFT
6. Updated the example to initialize the Xively Client *after* the device is connected to WiFi, so the SNTP process isn't started any earlier
7. Improve timing of the GPIO button debouncer
8. Fixed an 'unused variable' warning in the protobuf code that caused the libxively.a build to fail

[ JIRA ]
ESP32 SFT Implementation
https://jira.3amlabs.net/browse/XCL-3080

[ Reviewer ]
@atigyi 

[ QA ]
My device successfully updates its firmware. I've tried multiple times, disconnected the device while it downloaded the new image, and other manual testing. Everything seems to be working correctly.

Sometimes, the ESP32's API fails to write the last chunk in a firmware update. The issue can be resolved running `make flash_erase` and re-flashing the device.

[ Release Notes ]
ESP32 now supports firmware updates over MQTT using Xively's Secure File Transfer protocol.


examples/esp32/xively_demo/main/main.c
```diff
@@ -303,3 +348,46 @@ void xt_recv_mqtt_msg_callback( const xi_sub_call_params_t* const params )
...
+    }
+    ota_download_progress = ( ( chunk_size + offset ) * 100 ) / ota_download_file_size;
+    printf( "\n[%d%%] Firmware chunk of size [%d] written at offset [%d]",
+            ota_download_progress, chunk_size, offset );
+
+    { /* FW update light show */
+        for ( int i = 4; i > 0; i-- )
+        {
+            io_led_set( i % 2 );
+            vTaskDelay( 50 / portTICK_PERIOD_MS );
```
atigyi: whoa, spooky lights flashing :)

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
```diff
@@ -0,0 +1,307 @@
...
+    {
+        const esp_partition_t* next_partition = esp_ota_get_next_update_partition( NULL );
+
+        if ( NULL == next_partition )
+        {
+            xi_bsp_debug_logger( "Failed to get next OTA update partition" );
+            return XI_BSP_IO_FS_OPEN_ERROR;
+        }
+
+        esp_retv = esp_ota_begin( next_partition, size,
```
atigyi: wow, size is needed here too. POSIX doesn't, CC3200 does need the size here in the open file function, great to see this!!

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
```diff
@@ -0,0 +1,307 @@
...
+    return ret;
+}
+
+xi_bsp_io_fs_state_t
+xi_bsp_io_fs_open( const char* const resource_name,
+                   const size_t size,
+                   const xi_bsp_io_fs_open_flags_t open_flags,
+                   xi_bsp_io_fs_resource_handle_t* resource_handle_out )
+{
+    char filepath[XI_ESP32_MAX_FILENAME_LEN] = "";
```
atigyi: isn't = { 0 }; better?

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
```diff
@@ -0,0 +1,307 @@
...
+    {
+        xi_bsp_debug_logger( "Error: No FS read of FW files in ESP32's OTA API" );
+        return XI_BSP_IO_FS_NOT_IMPLEMENTED;
+    }
+    else
+    {
+        /* make an allocation for memory block */
+        if ( NULL == *buffer )
+        {
+            *buffer = ( uint8_t* )xi_bsp_mem_alloc( xi_bsp_io_fs_buffer_size );
```
atigyi: this memory chunk is never deallocated: leaks

atigyi: please check the xi_bsp_io_fs_cc3200.c's read function and the buffer used there. A local BSP IO FS static buffer is used there and returned. Since other BSPs are written so, libxively internals (in this case xi_sft_revision.c) assume no deallocation is needed. So just define here a local static read buffer array, like this:
static uint8_t read_buffer[ xi_bsp_io_fs_buffer_size];

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
```diff
@@ -0,0 +1,307 @@
...
+
+    if ( XI_BSP_IO_FS_INVALID_RESOURCE_HANDLE == resource_handle )
+    {
+        xi_bsp_debug_logger( "\nInvalid resource handle - Filesystem read failed" );
+        return XI_BSP_IO_FS_INVALID_PARAMETER;
+    }
+
+    if ( ( esp_ota_handle_t )resource_handle == open_firmware_bin_handle )
+    {
+        xi_bsp_debug_logger( "Error: No FS read of FW files in ESP32's OTA API" );
```
atigyi: 👍 fair enough

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
atigyi: this is just a validation of the BSP API is "correct" and the parameter is not only used in one case: CC3200 but in ESP32 too, probably this shows that simpler file systems need size prior open, ok, probably only I am happy about it :)

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
```diff
@@ -0,0 +1,307 @@
...
+    int ret = -1;
+    if ( !xi_bsp_fwu_is_this_firmware( resource_name ) )
+    {
+        xi_bsp_debug_logger( "FS Remove not implemented for FW images" );
+        return XI_BSP_IO_FS_NOT_IMPLEMENTED;
+    }
+    else
+    {
+        char filepath[64] = "";
+        snprintf( filepath, 64, "%s/%s", ESP32_FS_BASE_PATH, resource_name );
```
atigyi: is 64 enough?

src/import/protobuf-c/library/protobuf-c.c
```diff
@@ -1863,6 +1866,7 @@ repeated_field_pack_to_buffer(const ProtobufCFieldDescriptor *field,
 		rv += uint32_pack(payload_len, scratch + rv);
 		buffer->append(buffer, rv, scratch);
 		tmp = pack_buffer_packed_payload(field, count, array, buffer);
+		(void) tmp; /* Warning fix by Xively */
 		assert(tmp == payload_len);
```
atigyi: Is this only ESP32 specific problem? I guess yet.
Can't we solve this by defining assert something like this:
#define assert( x ) (void)x
? 
This way a third party code could stay intact and a version refresh wouldn't break the ESP32 break.

src/import/protobuf-c/library/protobuf-c.c
```diff
@@ -25,6 +25,9 @@
  * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * -----
+ * 1 line (1869) added by Xively to fix a build warning
```
atigyi: this is too fragile an edit will invalidate the line number

src/libxively/sft/xi_sft_logic.c
```diff
@@ -298,6 +298,7 @@ xi_sft_on_message( xi_sft_context_t* context, xi_control_message_t* sft_message_
                             XI_CONTROL_MESSAGE__SFT_FILE_STATUS_PHASE_PROCESSING,
                             chunk_handling_status_code );
 
+                        xi_debug_logger( "Error: File chunk handling failed" );
```
atigyi: 👍

src/libxively/sft/xi_sft_logic.c
```diff
@@ -346,6 +347,7 @@ xi_sft_on_message( xi_sft_context_t* context, xi_control_message_t* sft_message_
                         if ( XI_CONTROL_MESSAGE__SFT_FILE_STATUS_CODE_SUCCESS !=
                              checksum_status_code )
                         {
+                            xi_debug_logger( "Error: File checksum validation failed" );
```
atigyi: 👍

src/libxively/sft/xi_sft_revision.c
```diff
@@ -77,13 +77,13 @@ static xi_state_t xi_sft_revision_read_string_from_file( const char* const resou
...
     XI_CHECK_STATE( state );
 
-    xi_bsp_io_fs_close( resource_handle );
-
     /* allocate output buffer then copy revision buffer into output buffer */
     XI_ALLOC_BUFFER_AT( char, *string_read_out, buffer_size + 1, state );
 
     memcpy( *string_read_out, buffer, buffer_size );
 
+    xi_bsp_io_fs_close( resource_handle );
```
atigyi: my change sinks this more deeper: under the err_handling, but this can stay here, But during merge put this under the err_handling label pls, thx!

examples/esp32/xively_demo/main/main.c
```diff
@@ -31,19 +31,33 @@
   #define USER_CONFIG_XI_DEVICE_PWD "[SET YOUR XIVELY DEVICE PASSWORD HERE]"
 #endif
 
-#define XT_TASK_ESP_CORE    0 /* ESP32 core the XT task will be pinned to */
-#define XT_TASK_STACK_SIZE  36 * 1024
-#define GPIO_TASK_STACK_SIZE 2 * 1024
+#define XT_TASK_ESP_CORE 0 /* ESP32 core the XT task will be pinned to */
+#define XT_TASK_STACK_SIZE ( 1024 * 36 )
```
atigyi: braces are very welcomed here 👍

examples/esp32/xively_demo/main/main.c
```diff
@@ -303,3 +348,46 @@ void xt_recv_mqtt_msg_callback( const xi_sub_call_params_t* const params )
...
+            io_led_set( i % 2 );
+            vTaskDelay( 50 / portTICK_PERIOD_MS );
+        }
+    }
+}
+
+void esp32_xibsp_notify_update_applied()
+{
+    printf( "\nFirmware package download complete - Proceeding with reboot" );
+    /* xi_bsp_fwu_on_package_download_finished called from the fwu BSP */
```
atigyi: do we need this? couldn't you put this printf into the xi_bsp_fwu_on_package_download_finished function directly?

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
```diff
@@ -0,0 +1,307 @@
...
+    if ( ( cnd ) )                                                                       \
+    {                                                                                    \
+        ( s ) = ( e );                                                                   \
+        perror( "Xi FS BSP errno" );                                                     \
+        goto err_handling;                                                               \
+    }
+
+/* These functions are implemented in the application layer to report status */
+extern void esp32_xibsp_notify_update_started( const char* filename, size_t file_size );
+extern void esp32_xibsp_notify_chunk_written( size_t chunk_size, size_t offset );
```
atigyi: I like this solution: wiring out notification to application code to use the device specific LED from there. But what if somebody doesn't link this xilib version to the demo? These functions will be undefined symbols during link time. Still this is a good solution.
But what about solving the same problem with function pointers? This BSP IO FS for ESP32 may define here function pointer types and actual pointers defaulting them NULL. The application may fill in these function pointers in its main.c. Then this BSP code can check against NULL value, if it isn't NULL, then it calls the function, if NULL the BSP does not emit the notification.
This solution lets user link this libxively version to any other application code even which does not define these functions.

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
```diff
@@ -0,0 +1,307 @@
...
+
+/* The size of the buffer to be used for reads. */
+const size_t xi_bsp_io_fs_buffer_size = 1024;
+/* ESP32 Over The Air FW updates API handle */
+static esp_ota_handle_t open_firmware_bin_handle = 0;
+
+/* translates bsp errno errors to the xi_bsp_io_fs_state_t values */
+xi_bsp_io_fs_state_t xi_bsp_io_fs_posix_errno_2_xi_bsp_io_fs_state( int errno_value )
```
atigyi: ..._posix_errno_2_... does the ESP32 FAT API return posix error values? Is the mapping valid for ESP32 too?
If yes, then ok. If not, then update the mapping and rename the function pls.

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        esp32_xibsp_notify_update_started( resource_name, size );
+        open_firmware_bin_handle = *resource_handle_out;
+    }
+    else
+    {
+        snprintf( filepath, XI_ESP32_MAX_FILENAME_LEN, "%s/%s", ESP32_FS_BASE_PATH,
+                  resource_name );
+        fp = fopen( ( const char* )filepath,
```
atigyi: this is a POSIX fopen function. Where does this come? Does the ESP32 SDK provide it or did you implement it?

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
```diff
@@ -0,0 +1,307 @@
...
+}
+
+xi_bsp_io_fs_state_t
+xi_bsp_io_fs_write( const xi_bsp_io_fs_resource_handle_t resource_handle,
+                    const uint8_t* const buffer,
+                    const size_t buffer_size,
+                    const size_t offset,
+                    size_t* const bytes_written )
+{
+    ( void )offset;
```
atigyi: offset is used, this line can be nuked to orbit

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        FILE* fp = ( FILE* )resource_handle;
+
+        /* let's set the offset */
+        const int fop_ret = fseek( fp, offset, SEEK_SET );
+
+        /* if error on fseek check errno */
+        XI_BSP_IO_FS_CHECK_CND(
+            fop_ret != 0, xi_bsp_io_fs_posix_errno_2_xi_bsp_io_fs_state( errno ), ret );
```
atigyi: have you tested the non firmware file update?

atigyi:
I see three .bin files in the PR:
- bootloader.bin
- xively_demo.bin
- xively_demo_update.bin

I think I know what the 2nd and 3rd is, but what is 1st? Bootloader?
I ask this because I don't see what doest bootloader do on ESP32 if not the FW rollback. But FW rollback is not there, so why is there a bootloader.bin file?

examples/esp32/xively_demo/main/main.c
Palantir555: :ghost:

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
Palantir555: I'm happy about it too! :) It would suck to download the entire size of the partition, only to have it fail at the end because there's not enough space. Glad our BSP is flexible enough, and glad the SDK devs keep it in mind.

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
Palantir555: I usually go with `""` for consistency (all strings, empty or not, are initialized using `"`).

But if our style guideline is to use `= { 0 };`, I certainly prefer being consistent at that level. I'll change it.

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
Palantir555: ☠️ ☠️☠️

Nice catch, thanks for the heads-up. Fixed 👍

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
Palantir555: I think so. I'm gonna increase it to 128, just in case.

examples/esp32/xively_demo/main/main.c
Palantir555: I try to use only `xi_debug_logger/format` in the BSPs, never printf()s. I added this message so the user knows why the device is rebooting, regardless of their BSP_DEBUG configuration.

If you think it's making things more confusing, I'll be happy to get rid of it.

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
Palantir555: I have not validated the map entries one by one, but yes, the ESP32 uses standard POSIX errno values.

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
Palantir555: Yes, the ESP32 SDK provides a POSIX-compliant filesystem API.

The file paths are used to dictate the external memory device and filesystem partition where each file should be stored (SPI flash, SD card, etc.)

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
Palantir555: Good point... I've only tested it with `.xirev` files, not regular updates. I'm gonna test it right now....

src/import/protobuf-c/library/protobuf-c.c
Palantir555: What do you mean by 'an ESP32 specific file'? I've tried modifying xi_debug.h to `#define assert( e ) ( ( void )(e) )` instead of `#define assert( e ) ( ( void )0 )`, but that didn't fix the problem. I think the ESP's SDK may be overwriting the macro, but I'm not sure.

I'd also like to avoid modifying 3rd party code, but I don't think this particular change is worth the time. Let me know if you disagree.

src/import/protobuf-c/library/protobuf-c.c
Palantir555: Fixed

Palantir555:
@atigyi I think the bootloader's build configuration can be easily modified with the SDK's configuration tool. That means we can't rely on other users' bootloaders being built with compatible settings, or in the same address as specified by our partition table.

[This](https://developer.xively.com/docs/esp32#section-task-5-of-6-flashing-the-device) is the command used to flash the device -- It includes the bootloader's binary and address.

You reminded me to add the compiled partitions table to the repo, though. Forgot that one 😅

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
atigyi: ❗️ what about this?

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
atigyi: 👍

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
atigyi: ok

examples/esp32/xively_demo/main/main.c
atigyi: ok, we can keep this but make it used via function pointers as stated below in one of my suggestions.

examples/esp32/xively_demo/main/xively_task.c
```diff
@@ -275,6 +290,28 @@ int8_t xt_connect( void )
     return 0;
 }
 
+static int8_t xt_configure_fw_updates( void )
+{
+    const char** updateable_files = ( const char* [] ){"firmware.bin"};
+    const uint16_t files_num      = 1;
```
atigyi: const:  👍

src/bsp/platform/esp32/xi_bsp_io_fs_esp32.c
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    if ( ( esp_ota_handle_t )resource_handle == open_firmware_bin_handle )
+    {
+        xi_bsp_debug_logger( "Error: No FS read of FW files in ESP32's OTA API" );
+        return XI_BSP_IO_FS_NOT_IMPLEMENTED;
+    }
+    else
+    {
+        static uint8_t read_buffer[XI_BSP_IO_FS_READ_BUFFER_SIZE] = {0};
```
atigyi: 👍

src/import/protobuf-c/library/protobuf-c.c
```diff
@@ -46,7 +46,7 @@
  */
 
 #include <stdlib.h>	/* for malloc, free */
-#include <string.h>	/* for strcmp, strlen, memcpy, memmove, memset */
+#include <string.h>     /* for strcmp, strlen, memcpy, memmove, memset */
```
atigyi: LAST: pls remove this extra space to have no change at all in this 3rd party code, thx

Palantir555: Oopse! Fixed

src/bsp/platform/esp32/xi_bsp_sft_notifications_esp32.c
```diff
@@ -0,0 +1,8 @@
+/* Copyright (c) 2003-2017, LogMeIn, Inc. All rights reserved.
+ *
+ * This is part of the Xively C Client library,
+ * it is licensed under the BSD 3-Clause license.
+ */
```
atigyi: I'd rename this file to `xi_bsp_fwu_notifications_esp32.c`. Why? Because the FWU is already used as a BSP module. Even do this when this FWU notification is not only used from the FWU BSP but also from the IO FS BSP.

src/bsp/platform/esp32/include/xi_esp32_sft_notifications.h
```diff
@@ -0,0 +1,34 @@
+/* Copyright (c) 2003-2017, LogMeIn, Inc. All rights reserved.
+ *
```
atigyi: I'd rename this to `xi_bsp_fwu_notifications_esp32.h` or `xi_bsp_fwu_notifications.h`.

src/bsp/platform/esp32/include/xi_esp32_sft_notifications.h
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef __XI_ESP32_SFT_NOTIFICATIONS_H__
+#define __XI_ESP32_SFT_NOTIFICATIONS_H__
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+#include <stdio.h>
+
+typedef void ( xi_bsp_notif_update_started_callback_t )( const char* filename,
```
atigyi: I'd change the prefix of all these types to `xi_bsp_fwu_notifications_...`

src/bsp/platform/esp32/include/xi_esp32_sft_notifications.h
```diff
@@ -0,0 +1,34 @@
...
+typedef void ( xi_bsp_notif_chunk_written_callback_t )( size_t chunk_size,
+                                                        size_t offset );
+typedef void ( xi_bsp_notif_update_applied_callback_t )( void );
+
+typedef struct
+{
+    xi_bsp_notif_update_started_callback_t* update_started;
+    xi_bsp_notif_chunk_written_callback_t* chunk_written;
+    xi_bsp_notif_update_applied_callback_t* update_applied;
+} bsp_notification_callbacks_t;
```
atigyi: add `xi_...` prefix

Palantir555:
@atigyi I've implemented the changes you suggested. Please, approve for release

atigyi:
approved!! 👍 great job!!

PR #161, opened 2017-10-27
Feature/sft file ordering

DellaBitta:
[ Description ]
Added a BSP function to the FWU BSP to let the Application order the downloads of SFT resources.  This function, xi_bsp_io_fwu_order_resource_downloads(), will be invoked a FILE_UPDATE_AVAILABLE message is sent by the Xively SFT service.

[ JIRA ]
XCL-3081
https://jira.3amlabs.net/browse/XCL-3081

[ Reviewer ]

[ QA ]
Tested by altering the MQTT Logic Producer to download files on POSIX.  Changed the order in xi_bsp_io_fwu_order_resource_downloads to download in a random order of three files, and everything downloaded correctly.  Tested multiple runs to hit all combinations of file orders.  Built for POSIX and tested on POSXI.  Built on CC3200 and STM32F4 toolchains.

Ran code formatter on changed files

[ Release Notes ]
Added the ability for Client Applications to order the files downloaded over the Xively Secure File Transfer service.  Please see the documentation for xi_bsp_io_fwu_order_resource_downloads() in the header include/bsp/xi_bsp_fwu.h for more information.


include/bsp/xi_bsp_fwu.h
```diff
@@ -84,6 +84,23 @@ void xi_bsp_fwu_on_package_download_failure();
 
 /**
  * @function
+ * @brief Used by the Xively C Client to determine which resource to download
+ * next. This function is invoked ONLY when XI_SFT_BSP_CHOOSE_DOWNLOAD_ORDER
+ * is defined as a preprocessor definition when compiling the Xively C
+ * Client library.
+ *
```
DellaBitta: I thought that this function probably wouldn't be used on most platforms, so having it disabled by default could save some footprint, runtime memory usage, and complexity.  Thoughts?

examples/mqtt_logic_producer/src/mqtt_logic_producer.c
```diff
@@ -307,7 +306,7 @@ void delayed_publish( xi_context_handle_t context_handle,
     XI_UNUSED( user_data );
 
     printf( "publishing msg\n" );
-    
+
     /* sending the connect request */
     xi_publish( context_handle, xi_publishtopic, "Hello From Xively!", xi_example_qos,
                 XI_MQTT_RETAIN_FALSE, NULL, NULL );
```
DellaBitta: code formatter changes.

include/xively.h
```diff
@@ -163,12 +163,12 @@ extern xi_state_t xi_delete_context( xi_context_handle_t context_handle );
...
+ *
  * @retval 1 if the context is currently connected
  * @retval 0 if the context is invalid, or the connection is currently any
  * of the other following states: Unitialized, connecting, closing or closed.
  */
- extern uint8_t xi_is_context_connected( xi_context_handle_t context_handle );
+extern uint8_t xi_is_context_connected( xi_context_handle_t context_handle );
 
 
 /**
```
DellaBitta: code formatter changes

src/libxively/sft/xi_sft_logic_file_chunk_handlers.c
```diff
@@ -41,7 +41,7 @@ xi_sft_on_message_file_chunk_process_file_chunk( xi_sft_context_t* context,
     /* write bytes through FILE BSP */
     size_t bytes_written = 0;
 
-    state = xi_fs_bsp_io_fs_2_xi_state( 
+    state = xi_fs_bsp_io_fs_2_xi_state(
         xi_bsp_io_fs_write( context->update_file_handle, sft_message_in->file_chunk.chunk,
                             sft_message_in->file_chunk.length,
                             sft_message_in->file_chunk.offset, &bytes_written ) );
```
DellaBitta: code formatter.

include/bsp/xi_bsp_fwu.h
```diff
@@ -84,6 +84,22 @@ void xi_bsp_fwu_on_package_download_failure();
...
+ * @param [in] resource_names an array of resource name strings to choose from.
+ * @param [in] list_len the number of elements in the resource_names array.
+ *
+ * @return uint16_t the index in resource_names with the name of the
+ * resource to download next.
+ */
+uint16_t xi_bsp_io_fwu_get_next_resource_to_download( const char* const* resource_names,
```
atigyi: Is this called before each file download?

src/libxively/control_topic/xi_control_message.h
```diff
@@ -54,6 +54,8 @@ typedef struct xi_control_message_file_desc_ext_s
 
     uint8_t* fingerprint;
     uint16_t fingerprint_len;
+
+    uint8_t is_processed;
```
atigyi: This is a protocol descriptor file. Adding a local client state specific field to a SFT protocol message is a mixing of different areas: SFT protocol and the client's status of package download. Even if this info will be needed in the final solution it'd be better to store outside the client side protocol description.

atigyi:
I've put some notes under the "files changes" tab. This solution seems unnecessarily complex to me. A map is rebuilt each time before a file download. Also for simplicity reasons I think we can be brave and assume that the full file download order is known before the first file download. That would cause the libxively to be less paranoic and just ask the download order once at the beginning.
This solution would result in a single application call: reorder please. The default order array can be the increasing integer numbers, this can be reordered by the application's reorder-if-you-dare function. This order array may be stored in `xi_sft_context_t`. The function `xi_control_message_file_update_available_get_next_file_desc_ext` should be adapted to use this order array during the determination of the next file to download.
This would eliminate the need of the regeneration and deallocation of bsp-sft file index map at each file download.

include/bsp/xi_bsp_fwu.h
```diff
@@ -84,6 +84,28 @@ void xi_bsp_fwu_on_package_download_failure();
 
 /**
  * @function
+ * @brief Invoked at the start of a SFT download process, when SFT reports
+ * to the client that there are new resources to download.  This function
+ * lets the client application select the order in which resources will be
+ * downloadd.
```
atigyi: downloaded

include/bsp/xi_bsp_fwu.h
```diff
@@ -84,6 +84,28 @@ void xi_bsp_fwu_on_package_download_failure();
...
+ * @param [in] list_len the number of elements in the resource_names array.
+ * @param [in/out] download_order an integer array to be filled out by the
+ * BSP implementation.  The values in this array must be indicies of the
+ * resource_names array, ordered in the sequence that the resources should be
+ * downloaded. Values must be between 0 and list_len - 1.  When this function
+ * is invoked, this array is prepopulated with a default download order --
+ * this function may return immediately without altering the array to use
+ * these defaults.
+ */
+void xi_bsp_io_fwu_order_resource_downloads( const char* const* resource_names,
```
atigyi: xi_bsp_io_fwu_... ---> xi_bsp_fwu_...

src/libxively/sft/xi_sft_logic.c
```diff
@@ -8,6 +8,8 @@
 #include <stddef.h>
 #include <xi_macros.h>
 #include <xi_debug.h>
+#include <xi_vector.h>
+#include <xi_helpers.h>
```
atigyi: are these xi_vector.h and xi_helpers.h still needed?

src/libxively/sft/xi_sft_logic.c
```diff
@@ -18,6 +20,13 @@
 
 #include <xi_fs_bsp_to_xi_mapping.h>
 
+typedef struct xi_sft_remaining_resources_info_s
+{
+    const char** resource_names;
+    xi_vector_t* bsp_to_sft_index_map_vector;
+
+} xi_sft_remaining_resources_info_t;
+
```
atigyi: is this struct still needed?

src/libxively/sft/xi_sft_logic.c
```diff
@@ -51,6 +60,10 @@ xi_state_t xi_sft_make_context( xi_sft_context_t** context,
     return state;
 
 err_handling:
+    if ( NULL != *context )
+    {
+        XI_SAFE_FREE( ( *context )->updateable_files_download_order );
+    }
```
atigyi: I think this isn't necessary here, is it?

src/libxively/sft/xi_sft_logic.c
```diff
@@ -131,6 +144,89 @@ xi_sft_send_file_status( const xi_sft_context_t* context,
...
+    {
+        if ( 0 <= context->updateable_files_download_order[i] )
+        {
+            selected_index = context->updateable_files_download_order[i];
+            context->updateable_files_download_order[i] = -1;
+            break;
+        }
+    }
+
+    if ( selected_index >= context->update_message_fua->file_update_available.list_len )
```
atigyi: this is a check if the selected index is out of bounds, this check is already done in the other function, I will write comment there too. I think either of these two is enough.

src/libxively/sft/xi_sft_logic.c
```diff
@@ -131,6 +144,89 @@ xi_sft_send_file_status( const xi_sft_context_t* context,
...
+    {
+        return XI_ELEMENT_NOT_FOUND;
+    }
+
+    if ( -1 != selected_index )
+    {
+        printf( "Selected index: %d\n", selected_index );
```
atigyi: comment out or remove the printf line

src/libxively/sft/xi_sft_logic.c
```diff
@@ -131,6 +144,89 @@ xi_sft_send_file_status( const xi_sft_context_t* context,
...
+
+xi_state_t xi_sft_order_resource_downloads( xi_sft_context_t* context )
+{
+    xi_state_t state = XI_STATE_OK;
+    uint16_t i       = 0; /* forward declaration to suppress toolchain warnings */
+
+    const uint16_t num_incoming_resources =
+        context->update_message_fua->file_update_available.list_len;
+
+    context->updateable_files_download_order =
```
atigyi: from line 185-192 this could be solved with the internal allocators:
```

```

src/libxively/sft/xi_sft_logic.h
```diff
@@ -19,13 +19,15 @@ typedef struct
     void* send_message_user_data;
     const char** updateable_files;
     uint16_t updateable_files_count;
+    int32_t* updateable_files_download_order;
```
atigyi: I think int16_t would be enough, but this comment smells like nitpicking..., so I leave it up to you.

src/libxively/sft/xi_sft_logic.c
```diff
@@ -131,6 +144,89 @@ xi_sft_send_file_status( const xi_sft_context_t* context,
...
+            sizeof( char* ) * num_incoming_resources );
+    memset( resource_names, 0, num_incoming_resources * sizeof( char* ) );
+
+
+    for ( i = 0; i < num_incoming_resources; ++i )
+    {
+        context->updateable_files_download_order[i] = i;
+        resource_names[i] =
+            context->update_message_fua->file_update_available.list[i].name;
+        printf( "i: %d: %s\n", i, resource_names[i] );
```
atigyi: remove printf or comment it out, or make it debuglog

src/libxively/sft/xi_sft_logic.c
```diff
@@ -131,6 +144,89 @@ xi_sft_send_file_status( const xi_sft_context_t* context,
...
+    xi_bsp_io_fwu_order_resource_downloads( ( const char* const* )resource_names,
+                                            num_incoming_resources,
+                                            context->updateable_files_download_order );
+
+    XI_SAFE_FREE( resource_names );
+
+    /* Check for index out of bounds */
+    for ( i = 0; i < num_incoming_resources; ++i )
+    {
+        XI_CHECK_CND(
```
atigyi: here are the returning indices bound checked, I think it is enough to make out of bound check once

src/libxively/sft/xi_sft_logic.c
```diff
@@ -131,6 +144,89 @@ xi_sft_send_file_status( const xi_sft_context_t* context,
...
+        resource_names[i] =
+            context->update_message_fua->file_update_available.list[i].name;
+        printf( "i: %d: %s\n", i, resource_names[i] );
+    }
+
+    xi_bsp_io_fwu_order_resource_downloads( ( const char* const* )resource_names,
+                                            num_incoming_resources,
+                                            context->updateable_files_download_order );
+
+    XI_SAFE_FREE( resource_names );
```
atigyi: 👍

src/libxively/sft/xi_sft_logic.c
```diff
@@ -293,10 +390,7 @@ xi_sft_on_message( xi_sft_context_t* context, xi_control_message_t* sft_message_
 
                     /* continue the update package download  */
                     {
-                        context->update_current_file =
-                            xi_control_message_file_update_available_get_next_file_desc_ext(
```
atigyi: please also remove the header and implementation of this old function.

src/libxively/sft/xi_sft_logic.c
```diff
@@ -131,6 +144,89 @@ xi_sft_send_file_status( const xi_sft_context_t* context,
     ( *context->fn_send_message )( context->send_message_user_data, message_file_status );
 }
 
+xi_state_t xi_sft_select_next_resource_to_download( xi_sft_context_t* context )
```
atigyi: could you please add unit tests on this function. In the unit test case setups you could add any weird index map: out of bounds, readding the same index again, and/or leaving some indices out replacing them with redundant indices

atigyi: also I've started to add `_` (underscore) prefixes for .c private functions, also a `static` classifier. This `static` keyword makes the function not visible from outside of the compilation / translation unit. But yes, if you make this function static, then the unit test won't be able to re-use it in the test cases. So this `static` stuff does not play here but the `_` prefix still can stay in the game.

src/libxively/sft/xi_sft_logic.c
DellaBitta: I started with these, but the compiler produces warnings (and errors) that the buffers aren't always used due to a potential if path.   I'm not sure I understand why.

src/libxively/sft/xi_sft_logic.h
DellaBitta: yeah, using 32 just to be safe

src/libxively/sft/xi_sft_logic.c
DellaBitta: If there's an error in creating the context, the xi_sft_free_context is not called.  Therefore we should catch errors and remove any sort of allocations that were made in an attempt to build the context, no?  In this case this is only the first allocation in the context, but this is safer to do in case people add more allocations later, and it's harmless for now.

src/libxively/sft/xi_sft_logic.c
DellaBitta: Ok, removing this one.

src/libxively/sft/xi_sft_logic.c
DellaBitta: :+1:

src/libxively/sft/xi_sft_logic.c
DellaBitta: for instance, for the first one where we aqllocate the download order array:

/Users/ddellabitta/dev/xively/xively-client-c/src/libxively/sft/xi_sft_logic.c:173:5: error: variable 'resource_names' is used uninitialized whenever 'if' condition is true [-Werror,-Wsometimes-uninitialized]
    XI_ALLOC_BUFFER_AT( int32_t, context->updateable_files_download_order,
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/ddellabitta/dev/xively/xively-client-c/src/libxively/xi_macros.h:83:5: note: expanded from macro 'XI_ALLOC_BUFFER_AT'
    XI_CHECK_MEMORY( out, state );                                                       \
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/ddellabitta/dev/xively/xively-client-c/src/libxively/xi_macros.h:65:33: note: expanded from macro 'XI_CHECK_MEMORY'
#define XI_CHECK_MEMORY( a, s ) XI_CHECK_CND( ( a ) == 0, XI_OUT_OF_MEMORY, s )
                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/ddellabitta/dev/xively/xively-client-c/src/libxively/xi_macros.h:30:10: note: expanded from macro 'XI_CHECK_CND'
    if ( ( cnd ) )                                                                       \
         ^~~~~~~
/Users/ddellabitta/dev/xively/xively-client-c/src/libxively/sft/xi_sft_logic.c:215:19: note: uninitialized use occurs here
    XI_SAFE_FREE( resource_names );
                  ^~~~~~~~~~~~~~
/Users/ddellabitta/dev/xively/xively-client-c/src/libxively/xi_macros.h:59:12: note: expanded from macro 'XI_SAFE_FREE'
    if ( ( a ) )                                                                         \
           ^
/Users/ddellabitta/dev/xively/xively-client-c/src/libxively/sft/xi_sft_logic.c:173:5: note: remove the 'if' if its condition is always false
    XI_ALLOC_BUFFER_AT( int32_t, context->updateable_files_download_order,
    ^
/Users/ddellabitta/dev/xively/xively-client-c/src/libxively/xi_macros.h:83:5: note: expanded from macro 'XI_ALLOC_BUFFER_AT'
    XI_CHECK_MEMORY( out, state );                                                       \
    ^
/Users/ddellabitta/dev/xively/xively-client-c/src/libxively/xi_macros.h:65:33: note: expanded from macro 'XI_CHECK_MEMORY'
#define XI_CHECK_MEMORY( a, s ) XI_CHECK_CND( ( a ) == 0, XI_OUT_OF_MEMORY, s )
                                ^
/Users/ddellabitta/dev/xively/xively-client-c/src/libxively/xi_macros.h:30:5: note: expanded from macro 'XI_CHECK_CND'
    if ( ( cnd ) )                                                                       \
    ^
/Users/ddellabitta/dev/xively/xively-client-c/src/libxively/sft/xi_sft_logic.c:176:9: note: variable 'resource_names' is declared here
        char** resource_names = xi_alloc( num_incoming_resources * sizeof( char* ) );
        ^
1 error generated.

src/libxively/sft/xi_sft_logic.c
DellaBitta: :+1:

src/libxively/sft/xi_sft_logic.c
DellaBitta: :+1:

src/libxively/sft/xi_sft_logic.c
atigyi: I get confused by this, because I check from where can the execution jump to the err_handling label. And the only place is the XI_ALLOC_AT. In case the execution jumps from there then the context will be null. So this code never gets called. And there is countless code blocks we could introduce which won't be executed ever. So what's bugging me is why have you chosen this code from the infinite number of possible code blocks not to be executed ever.

src/libxively/sft/xi_sft_logic.c
atigyi: true, then use this:

```
char** resource_names = NULL;

    XI_ALLOC_BUFFER_AT( int32_t, context->updateable_files_download_order,
                        sizeof( int32_t ) * num_incoming_resources, state );
    XI_ALLOC_BUFFER_AT( char*, resource_names, sizeof( char* ) * num_incoming_resources,
                        state );
```

atigyi: or the order change solves the issue too:
```
    XI_ALLOC_BUFFER( char*, resource_names, sizeof( char* ) * num_incoming_resources,
                     state );
    XI_ALLOC_BUFFER_AT( int32_t, context->updateable_files_download_order,
                        sizeof( int32_t ) * num_incoming_resources, state );
```

src/tests/utests/xi_utest_sft_logic.c
```diff
@@ -231,6 +231,292 @@ XI_TT_TESTCASE_WITH_SETUP(
...
+                           } )
+
+XI_TT_TESTCASE_WITH_SETUP(
+    xi_utest__select_next_resource_to_download__picks_first_element,
+    xi_utest_setup_basic,
+    xi_utest_teardown_basic,
+    NULL,
+    {
+        xi_control_message_file_desc_ext_t three_file_list[3];
+        memset( three_file_list, 0, 3 * sizeof( xi_control_message_file_desc_ext_t ) );
```
atigyi: this one liner does the same as the two here:
`xi_control_message_file_desc_ext_t three_file_list[3] = {0};`
https://stackoverflow.com/questions/2589749/how-to-initialize-array-to-0-in-c

src/tests/utests/xi_utest_sft_logic.c
```diff
@@ -231,6 +231,292 @@ XI_TT_TESTCASE_WITH_SETUP(
...
 
+XI_TT_TESTCASE_WITH_SETUP( xi_utest__select_next_resource_to_download__null_context,
+                           xi_utest_setup_basic,
+                           xi_utest_teardown_basic,
+                           NULL,
+                           {
+                               xi_state_t state =
+                                   _xi_sft_select_next_resource_to_download( NULL );
+
+                               tt_want_int_op( XI_INVALID_PARAMETER, ==, state );
```
atigyi: 👍

src/tests/utests/xi_utest_sft_logic.c
```diff
@@ -231,6 +231,292 @@ XI_TT_TESTCASE_WITH_SETUP(
...
+
+        int32_t download_order[] = {0, 1, 2};
+        xi_sft_context_t sft_context;
+        sft_context.updateable_files_download_order = download_order;
+        sft_context.update_current_file             = NULL;
+        sft_context.update_message_fua              = &file_update_available;
+
+        xi_state_t state = _xi_sft_select_next_resource_to_download( &sft_context );
+        tt_want_int_op( XI_STATE_OK, ==, state );
+        tt_want_ptr_op( sft_context.update_current_file, ==, &three_file_list[0] );
```
atigyi: 👍

src/tests/utests/xi_utest_sft_logic.c
```diff
@@ -231,6 +231,292 @@ XI_TT_TESTCASE_WITH_SETUP(
...
+
+        int32_t download_order[] = {1, 1, 1};
+        xi_sft_context_t sft_context;
+        sft_context.updateable_files_download_order = download_order;
+        sft_context.update_current_file             = NULL;
+        sft_context.update_message_fua              = &file_update_available;
+
+        xi_state_t state = _xi_sft_select_next_resource_to_download( &sft_context );
+        tt_want_int_op( XI_STATE_OK, ==, state );
+        tt_want_ptr_op( sft_context.update_current_file, ==, &three_file_list[1] );
```
atigyi: hm, why is the second selected?

src/tests/utests/xi_utest_sft_logic.c
```diff
@@ -231,6 +231,292 @@ XI_TT_TESTCASE_WITH_SETUP(
...
+        download_order[127] = 0;
+
+        xi_sft_context_t sft_context;
+        sft_context.updateable_files_download_order = download_order;
+        sft_context.update_current_file             = NULL;
+        sft_context.update_message_fua              = &file_update_available;
+
+        xi_state_t state = _xi_sft_select_next_resource_to_download( &sft_context );
+        tt_want_int_op( XI_STATE_OK, ==, state );
+        tt_want_ptr_op( sft_context.update_current_file, ==, &large_file_list[0] );
```
atigyi: shouldn't this be &large_file_list[127] ?

src/tests/utests/xi_utest_sft_logic.c
DellaBitta: cool cool.

DellaBitta: unfortunately the compiler does not like this. :(

src/tests/utests/xi_utest_sft_logic.c
```diff
@@ -231,6 +231,292 @@ XI_TT_TESTCASE_WITH_SETUP(
...
+                           .msgver  = 1},
+                .list_len = 128,
+                .list     = large_file_list}};
+
+        int32_t download_order[128];
+        uint16_t i = 0;
+
+        for ( ; i < 127; ++i )
+        {
+            download_order[i] = -1;
```
atigyi: I would add few more unit tests with defining download order in the following ways:
```
download_order[i] = 127 - i;
download_order[i] = (i < 64) ? 126 - 2 * i : 2 * i - 127; // this is a V, zero at i==63, left side even, right side odd numbers, pick order should be: 63, 64, 62, 65, 61, 66, 60, 67, ... , 2, 125, 1, 126, 0, 127 
download_order[i] = (i < 64) ? 126 - 2 * i : 255 - 2 * i); // this is two backslashes, 126 at i==0, 0 at i==63, 127 at i==64, 1 at i==127, expected order: 63, 127, 62, 126, 61, 125, 60, 124, ... , 1, 65, 0, 64
``` 
I would do multiple `_xi_sft_select_next_resource_to_download` calls in the last unit tests and check all return values of the `_xi_sft_select_next_resource_to_download` calls to the end. 128 calls.

src/tests/utests/xi_utest_sft_logic.c
```diff
@@ -231,6 +231,292 @@ XI_TT_TESTCASE_WITH_SETUP(
...
+    } )
+
+XI_TT_TESTCASE_WITH_SETUP(
+    xi_utest__select_next_resource_to_download__negative_index_in_download_order,
+    xi_utest_setup_basic,
+    xi_utest_teardown_basic,
+    NULL,
+    {
+        // ARRANGE
+        xi_control_message_file_desc_ext_t three_file_list[3];
```
atigyi: maybe you could cut this arrange to a separate function since these lines appear multiple times in most of the test cases. but I leave it up to you, both have pros and cons.

src/tests/utests/xi_utest_sft_logic.c
atigyi: why, not first or last?

include/bsp/xi_bsp_fwu.h
```diff
@@ -84,6 +84,28 @@ void xi_bsp_fwu_on_package_download_failure();
...
+ * @param [in] resource_names an array of resource name strings that have
+ * new revisions in the SFT download package.
+ * @param [in] list_len the number of elements in the resource_names array.
+ * @param [in/out] download_order an integer array to be filled out by the
+ * BSP implementation.  The values in this array must be indicies of the
+ * resource_names array, ordered in the sequence that the resources should be
+ * downloaded. Values must be between 0 and list_len - 1.  When this function
+ * is invoked, this array is prepopulated with a default download order --
+ * this function may return immediately without altering the array to use
+ * these defaults.
```
atigyi: I'd add here a small example for explanation:
`For a list of 4 the default order is: [0, 1, 2, 3]. The application might alter this and return and array like this: [2, 3, 1, 0]`. This says a lot, at least for me, that this is not a rocket science, just reorder if you want.

src/tests/utests/xi_utest_sft_logic.c
DellaBitta: because the first index in the download order array points to the second file in the file list.  This ensures that we iteratively go left to right in the list and find the first file that's available.  Even if the index is repeated, as this function doesn't guarantee index uniqueness.

src/tests/utests/xi_utest_sft_logic.c
DellaBitta: No, the last entry in the download index array is 0, which means the first file in the large_file_list.

src/tests/utests/xi_utest_sft_logic.c
atigyi: ah, right

src/tests/utests/xi_utest_sft_logic.c
atigyi: 👍

src/tests/utests/xi_utest_sft_logic.c
atigyi: right, since the members aren't primitive type but struct, so ok, memset is the tool to zero it out 👍

include/bsp/xi_bsp_fwu.h
DellaBitta: :+1:

src/libxively/sft/xi_sft_logic.c
DellaBitta: blort, nice. Thanks!
:+1:

atigyi:
Approved!

PR #162, opened 2017-10-30
Feature/sft qa autotests

atigyi:
[ Description ]
Fixing possible memory leak and not closed file in case SFT package download breaks because socket disconnection. Unit tests were introduced to reproduce in-SFT-process deallocation of SFT logic and checking deallocations afterwards.

[ JIRA ]
SFT feature fix
link: n/a

[ Reviewer ]

[ QA ]
New unit tests were written to cover SFT logic deallocation in the middle of a package download.

[ Release Notes ]
n/a

### DIFF
--- a/examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
+++ b/examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
@@ -1043,7 +1043,7 @@ void xi_parse_file_chunk( xi_context_handle_t in_context_handle,
 void verify_sha256( xi_context_handle_t in_context_handle )
 {
     /* Finalize SHA256 Digest */
-    xi_bsp_fwu_checksum_final( download_ctx.checksum_context,
+    xi_bsp_fwu_checksum_final( &download_ctx.checksum_context,
                                &download_ctx.file_local_computed_fingerprint,
                                &download_ctx.file_local_computed_fingerprint_len );
     printf( "Calculated hash = 0x" );
--- a/src/libxively/sft/xi_sft_logic.c
+++ b/src/libxively/sft/xi_sft_logic.c
@@ -60,6 +60,10 @@ xi_state_t xi_sft_free_context( xi_sft_context_t** context )
 {
     if ( NULL != context && NULL != *context )
     {
+        xi_sft_on_message_file_chunk_checksum_final( *context );
+
+        xi_bsp_io_fs_close( ( *context )->update_file_handle );
+
         xi_control_message_free( &( *context )->update_message_fua );
 
         XI_SAFE_FREE( *context );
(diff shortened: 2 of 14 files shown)


src/libxively/sft/xi_sft_logic.c
```diff
@@ -60,6 +60,10 @@ xi_state_t xi_sft_free_context( xi_sft_context_t** context )
 {
     if ( NULL != context && NULL != *context )
     {
+        xi_sft_on_message_file_chunk_checksum_final( *context );
+
+        xi_bsp_io_fs_close( ( *context )->update_file_handle );
```
DellaBitta: Wonder if we should pass an error to this function in case someone wants to roll back the file, if possible.

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
```diff
@@ -1043,7 +1043,7 @@ void xi_parse_file_chunk( xi_context_handle_t in_context_handle,
 void verify_sha256( xi_context_handle_t in_context_handle )
 {
     /* Finalize SHA256 Digest */
-    xi_bsp_fwu_checksum_final( download_ctx.checksum_context,
+    xi_bsp_fwu_checksum_final( &download_ctx.checksum_context,
```
DellaBitta: is this due to the upcoming HTTP changes?

src/libxively/sft/xi_sft_logic.c
atigyi: ? You mean "error, file close requested since there was an error (socket disconnection) during file content processing" or something like this?
And here the application may continue download from this point?
If this is the idea, I'd say this is too advanced subfeature for our current status: undestaffed + this SFT feature isn't sold at all yet. And nobody requested this feature.

examples/cc3200/xively_firmware_updates/SFT/xi_sft_client.c
atigyi: no. The checksum context deallocation is the responsibility of the `...final` function. And there is a style in the library that the deallocator does the NULL-ing the pointer itself. And this caused problem during destruction of SFT context: the checksum context was re-deallocated :D because the pointer wasn't nullified at the "normal" deallocation.

atigyi:
thx, well spotted, Travis build failed because of a bug in BSP checksum cryptoalg impl. Thx!!

src/libxively/sft/xi_sft_logic.c
DellaBitta: I wouldn't say "download again later" but perhaps "Not commit this file change, or Roll it back"
I could see some systems that don't overwrite files, but write to a temporary file until the checksum calculation occurs, and then moves the file.

PR #163, opened 2017-11-09
Adds SFT OTA to the CC3220SF

Rhunter1:
[ Description ]

Working OTA firmware updates for the CC3220SF using SFT:

1. Implemented the CC3220SF's file and fwu BSPs
2. Updated the example to set the updateble files.
3. Updated the example to include an OTA configuration.
4. Updated both CC3220SF presets to enable the control topic and SFT.

[ JIRA ]

[ Reviewer ]

[ QA ]
My device successfully updates its firmware with both the embedded TLS and wolfSSL TLS examples. I've tried multiple times as well as did multiple successful rollbacks on loading a "failing" FWU.

[ Release Notes ]
Built with the following tools:
- CCS v7.3.0
- ti-cgt-arm_16.9.4.LTS
- SimpleLink CC32XX SDK v1.50.00.06
- xdctools v3.50.03.33

An additional uniflash configuration was added to _examples/cc3220sf/xively_demo_ to generate the OTA tar package. The static files (i.e. xively_cfg.txt) should not be included in this uniflash configuration.
This package has not been tested with forcing the OTA package to be signed.
- To test this change the define **OTA_FORCE_SIGNATURE_VERIFICATION** to **TRUE** in the _OtaArchive.c_ file in the XivelyExample project.


make/mt-config/mt-presets.mk
```diff
@@ -22,8 +22,8 @@ CONFIG_CC3200                   =memory_fs-control_topic-tls_bsp-secure_file_tra
 CONFIG_CC3200_TLS_SOCKET        =memory_fs-control_topic-tls_socket-secure_file_transfer
 
 # CONFIG for CC3220
-CONFIG_CC3220SF                 =bsp_cc3220sf-memory_fs-tls_bsp
-CONFIG_CC3220SF_TLS_SOCKET      =bsp_cc3220sf-memory_fs-tls_socket
+CONFIG_CC3220SF                 =bsp_cc3220sf-memory_fs-control_topic-tls_bsp_socket-secure_file_transfer
```
atigyi: 👍 great update even in mt-presets.mk grats!

tls_bsp_socket --> tls_bsp
tls_bsp_socket works too since in mt-config.mk the string "tls_bsp" is searched for

src/bsp/platform/cc3220sf/xi_bsp_fwu_cc3220sf.c
```diff
@@ -0,0 +1,143 @@
...
+        xi_bsp_debug_format( "Error attempting a rollback, status: [ %i ]", status );
+    }
+    return status;	
+}
+
+
+
+uint8_t xi_bsp_fwu_is_this_firmware( const char* const resource_name )
+{
+    return ( 0 == strcmp( "firmware.tar", resource_name ) ) ? 1 : 0;
```
atigyi: 👍

src/bsp/platform/cc3220sf/xi_bsp_fwu_cc3220sf.c
```diff
@@ -0,0 +1,143 @@
...
+}
+
+xi_bsp_fwu_state_t xi_bsp_fwu_on_new_firmware_ok()
+{
+    if ( _ota_get_pending_commit() )
+    {
+		xi_bsp_debug_logger(" Pending commit, firmware OK." );
+        
+		/* What should we do if we get a failure to commit? Reboot? */
```
atigyi: hm, great question. commit failure means the device couldn't set the current FW to the default one, which means after next reboot not this but the previous FW will start. Right?! If we rebooted here it'd just shorten this new FW's runtime and fallback immediately to previous FW.
I don't know, should we retry _ota_commit() later? Schedule a job for this? From this BSP function the client's scheduler API isn't accessible though.

Rhunter1: For the CC3220SF there is a bootloader watchdog timer that is set when the firmware tar file is being accessed only. If the commit fails I believe that this watchdog timer will elapse and force a rollback. Does the watchdog timer need to be set on all updates including ones that do not update the firmware?

PR #169, opened 2017-12-20
CBOR Assertions Now Only in Debug Mode

DellaBitta:
[ Description ]
cn-cbor contains assertions that were active in the release builds of clients. These are now only on when XI_DEBUG_ASSERT is 1 in the build environment.  This is generally only toggled on during debug builds.

[ JIRA ]
None.

[ Reviewer ]
@atigyi

[ QA ]
Built and ran tests with debug and release mode.
added #error test to ensure that the debug assertion preprocessor definitions was toggled on in debug mode.

[ Release Notes ]
Assertions in CBOR encoder have been disabled in release mode.

### DIFF
--- a/src/import/cn-cbor/src/cn-cbor.c
+++ b/src/import/cn-cbor/src/cn-cbor.c
@@ -11,7 +11,11 @@ extern "C" {
 #include <stdlib.h>
 #include <stdint.h>
 #include <string.h>
-#include <assert.h>
+
+#if XI_DEBUG_ASSERT
+  #include <assert.h>
+#endif 
+
 #include <math.h>
 // #include <arpa/inet.h> // needed for ntohl (e.g.) on Linux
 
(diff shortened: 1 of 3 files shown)


src/import/cn-cbor/src/cn-cbor.c
```diff
@@ -11,7 +11,11 @@ extern "C" {
 #include <stdlib.h>
 #include <stdint.h>
 #include <string.h>
-#include <assert.h>
+
+#if XI_DEBUG_ASSERT
+  #include <assert.h>
+#endif
```
atigyi: I'd do:
```
#else
#define assert(X)
#endif
```
With this you don't need to do any changes inside the functions.

atigyi:
Additionally, I also see `#include <assert.h>` and `assert(...)` in file `cn-encoder.c` too. Isn't this problem too?!

PR #172, opened 2018-01-08
Fix ESP32/CC3200 SFT bug, and ESP32 demo issues

Palantir555:
[ Description ]

- Revert ESP32 tested SDK version to `e6afe28`, fix the libxively build for that PRESET, and remove the now-unnecessary lwipopts.h from the BSP
- Fix int overflow in `xi_bsp_time_getcurrenttime_milliseconds`
- Do not switch firmware images after SFT, if a firmware file was not part of the package 
- Implement light show to describe device connectivity at all times, and improve the existing ones for better consistency
- Miscellaneous clean-up

[ JIRA ]

[ Reviewer ]

[ QA ]
Built and run the code.

[ Release Notes ]
Fix int overflow in ESP32 time BSP.
Fix FWU BSPs on ESP32 and CC3200 for SFT packages without firmware binaries. Only commit/reboot if a new firmware image was downloaded.
Improve LED signalling on ESP32 demo.


Palantir555:
This PR has ended up collecting a bunch of fixes and improvements together. I can take the time to split them into different PR's, if you think it's worth the effort. Let me know.

examples/esp32/xively_demo/main/main.c
```diff
@@ -289,7 +287,7 @@ int8_t app_fetch_user_config( user_data_t* dst )
 
     /* Wait for a button press for a few seconds while flashing the LED. If the button is
      * pressed, clear the contents of user_config to force the provisioning process */
-    for ( int i = 40; i > 0; i-- )
+    for ( int i = 39; i >= 0; i-- )
```
DellaBitta: What's with the magic number here?

DellaBitta: Can it be defined at the top as a constant / Define?  
Or is there a platform system way to correlate ticks to time?

DellaBitta: Oh, and add to a ReadMe file somewhere in the project that you're concerned that the device reboots itself when the xively process crashes, and that people should think of another solution for their product . Then direct them to the code in the file for more information.

DellaBitta: If the comment is only in the code then it's likely to remain hidden from users.

src/bsp/platform/cc3200/xi_bsp_fwu_cc3200.c
```diff
@@ -76,7 +76,10 @@ void xi_bsp_fwu_order_resource_downloads( const char* const* resource_names,
 
 void xi_bsp_fwu_on_package_download_finished( const char* const firmware_resource_name )
 {
-    ( void )firmware_resource_name;
+    if ( NULL == firmware_resource_name )
```
atigyi: JC, sorry for nitpicking here, but I would do the opposite:
if ( NULL != firmware_resource_name)
{
include here the cc3200 firmwareupdate / reboot specific code
}

I think without any comment this code above tells more the intent: if there is a firmware_resource_name partition switch and reboot is necessary.
The current version tells me this: if there is no firmware_resource_name then there is nothing to do (even if this is true now it not necessary the case, one may do something with the downloaded package here).

And yes, nitpicking, sorry.

src/bsp/platform/esp32/xi_bsp_time_esp32.c
```diff
@@ -56,21 +55,23 @@ void xi_bsp_time_init()
...
 exit:
     sntp_stop();
 }
 
 xi_time_t xi_bsp_time_getcurrenttime_seconds()
 {
-    time_t current_time = 0;
-    XI_ESP32_GET_TIME_FROM_RTC( &current_time );
+    xi_time_t current_time = 0;
+    XI_ESP32_GET_TIME_FROM_RTC( ( time_t* )&current_time );
```
atigyi: I guess you don't need the time_t* cast anymore here.

atigyi:
Minor notes from my side, otherwise 👍

src/bsp/platform/cc3200/xi_bsp_fwu_cc3200.c
Palantir555: Sure, no problem, change implemented :)

src/bsp/platform/esp32/xi_bsp_time_esp32.c
Palantir555: Good point :+1:

Palantir555:
As discussed in standup, I'm ignoring the fuzz test error for this build (caused by external issues), and merging anyway.

PR #177, opened 2018-01-15
Fix fuzztest fail on travis

atigyi:
[ Description ]
There is a fuzz test target which gets built and run by Mr. Travis CI. The execution of this build failed before, with this change it works again.
Details: during it's run libfuzzer uses ptrace to monitor different aspects of the instrumented library / application, from a different thread I guess, maybe different process. Examining non-child processes or threads are disabled in linux kernel beginning with kernel version 10.10: https://wiki.ubuntu.com/SecurityTeam/Roadmap/KernelHardening#ptrace_Protection

Perfect solution would be changing libfuzzer to spawn System Under Test thread from the ptrace thread but I haven't chosen this. There is a fast workaround described in the link above: re-enabling user space processes to ptrace each other.

Also special, github version of clang is not needed anymore since recent clang releases already contain instrumentation flags required by libfuzzer. Also macOS' llvm compiler is able to compile with fuzzer flags, so fuzzing is also available on macOS. So this part is removed from the fuzztest makefile.

[ JIRA ]
n/a

[ Reviewer ]
@Palantir555

[ QA ]
Travis is able to run the fuzz test binary without an issue.

[ Release Notes ]
n/a

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: true
 language: c
 compiler:
   - clang
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,8 @@ env:
     - XI_BSP_TLS=wolfssl
     - XI_MAKETARGET=tests
   matrix:
-    - PRESET=ARM_REL
-    - PRESET=ARM_REL_MIN
+    # - PRESET=ARM_REL
+    # - PRESET=ARM_REL_MIN
 
     # - PRESET=STM32FX
     #   XI_MAKETARGET=""
@@ -24,20 +24,20 @@ env:
     # - PRESET=CC3200_TLS_SOCKET
     #   XI_MAKETARGET=""
 
-    - PRESET=ESP32
-      XI_MAKETARGET=""
+    # - PRESET=ESP32
+    #   XI_MAKETARGET=""
 
-    - PRESET=POSIX_DEV
-    - PRESET=POSIX_DEV_MIN
-    - PRESET=POSIX_REL
-    - PRESET=POSIX_REL_MIN
+    # - PRESET=POSIX_DEV
+    # - PRESET=POSIX_DEV_MIN
+    # - PRESET=POSIX_REL
+    # - PRESET=POSIX_REL_MIN
 
-    - PRESET=POSIX_REL
-      XI_BSP_TLS=mbedtls
+    # - PRESET=POSIX_REL
+    #   XI_BSP_TLS=mbedtls
 
-    - PRESET=POSIX_THREADING_REL
-    - PRESET=POSIX_UNSECURE_REL
-      XI_BSP_TLS=""
+    # - PRESET=POSIX_THREADING_REL
+    # - PRESET=POSIX_UNSECURE_REL
+    #   XI_BSP_TLS=""
 
     - TESTS=fuzztests
       PRESET=FUZZ_TESTS
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,13 +75,13 @@ matrix:
       env: PRESET=CC3200_TLS_SOCKET XI_MAKETARGET=""
     - compiler: clang
       env: PRESET=ESP32 XI_MAKETARGET=""
-    - compiler: gcc
+    - compiler: clang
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -76,12 +78,13 @@ matrix:
     - compiler: clang
       env: PRESET=ESP32 XI_MAKETARGET=""
     - compiler: gcc
-      env: TESTS=fuzztests PRESET=FUZZ_TESTS XI_BSP_TLS="" XI_MAKETARGET="fuzz_tests"
+      env: TESTS=fuzztests PRESET=FUZZ_TESTS XI_BSP_TLS="" XI_MAKETARGET="fuzz_tests" CXX=clang++
```
Palantir555: :+1:

.travis.yml
```diff
@@ -76,12 +78,13 @@ matrix:
...
 before_script:
+  # https://wiki.ubuntu.com/SecurityTeam/Roadmap/KernelHardening#ptrace_Protection
   - "if [[ ! $TESTS == 'fuzztests' ]]; then
         if [[ $CC == *'clang'* ]]; then export CC='clang-3.8'; fi;
      else
-        if [[ $CC == *'clang'* ]]; then export CC=clang CXX=clang++; fi;
+        echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope;
```
Palantir555: If I build the tests on my personal computer, will they permanently remove the `ptrace` protection?
If that's the case, we should figure out how to set it back to the original value `after_script`?

atigyi: this affects only Travis temporary virtual linux machines. These machines will be destroyed afterwards.
Local macos dev machines aren't affected.
`.travis.yml` is used only by travis.

Palantir555: d'oh, of course 😅  Thanks for clearing that up. PR Approved :+1:

Makefile
```diff
@@ -196,13 +196,12 @@ $(XI_FUZZ_TESTS_BINDIR)/%: $(XI_FUZZ_TESTS_SOURCE_DIR)/%.cpp
 	$(info [$(CXX)] $@)
 	$(MD) $(CXX) $< $(XI_CONFIG_FLAGS) $(XI_INCLUDE_FLAGS) -L$(XI_BINDIR) -L$(XI_LIBFUZZER_DOWNLOAD_DIR) $(XI_LIB_FLAGS) $(XI_FUZZ_TEST_LIBRARY) $(XI_COMPILER_OUTPUT)
 
-.PHONY: fuzz_tests
+$(XI_FUZZ_TESTS): $(XI)
```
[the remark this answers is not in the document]
atigyi: yes, I was thinking of this version. I even made the change you suggested. But I've decided to revert it. Why? I think this models more the reality. The fuzz test binaries require directly to be linked to libxively.a. The they are directly dependent from the $(XI). The `fuzz_tests` target actually executes the binaries. It depends directly from the binaries. So I think is more like a 2 level dependency and not a single level one.
But yes your version would make the same build effect, but we should put $(XI) before the $(XI_FUZZ_TESTS) which is fragile. And really shows that $(XI_FUZZ_TESTS) depends on $(XI) which is written now :)

make/mt-config/tests/mt-tests-fuzz.mk
```diff
@@ -57,13 +31,8 @@ $(XI_LIBFUZZER_DOWNLOAD_DIR):
...
+$(XI_LIBFUZZER): $(XI_LIBFUZZER_DOWNLOAD_DIR)
 	(cd $(XI_LIBFUZZER_DOWNLOAD_DIR) && clang++ -c -g -O2 -lstdc++ -std=c++11 *.cpp -IFuzzer && ar ruv libFuzzer.a Fuzzer*.o)
 
 $(XI_FUZZ_TESTS_CORPUS_DIRS):
 	@-mkdir -p $@
-
-#### =========================================================
-
-
-build_libfuzzer: $(XI_LIBFUZZER)
```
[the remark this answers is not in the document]
atigyi: yes it wasn't used at all. It was just a bias for understanding.

atigyi:
I think letting all processes `ptrace` eachother for temporary linux build machines isn't a security issue. It was the case for all linux version before 10 anyway.

Makefile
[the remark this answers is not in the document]
Palantir555: Fair enough :+1: Thanks

Issue #178, opened 2018-01-16
Update WolfSSL from 3.10.2 → 3.13.0 for Mac OSX


wcatron:
3.10.2 is throwing this error on `make`:
```
/Applications/Xcode.app/Contents/Developer/usr/bin/make -j9  all-am
  CCLD     src/libwolfssl.la
  CC       wolfcrypt/benchmark/benchmark.o
wolfcrypt/benchmark/benchmark.c:240:41: error: this function declaration is not a prototype [-Werror,-Wstrict-prototypes]
    WOLFSSL_API int wolfSSL_Debugging_ON();
                                        ^
                                         void
1 error generated.
make[1]: *** [wolfcrypt/benchmark/benchmark.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [all] Error 2
```

Which stops the initial xively-client-c `make` from properly installing WolfSSL. I ran `bash download_and_compile_wolfssl.sh` from the `src/imports/tls/` directory and got the same error.

---------

However once I updated WolfSSL to the latest stable version, WolfSSL compiled and installed properly but xively is now failing with a new error, probably from new configuration options with latest version.

```
# Using build PRESET: "POSIX_REL" to set CONFIG and TARGET variables:
.    XI_BSP_PLATFORM: [posix] 
.    XI_BSP_TLS:      [wolfssl] 

# Xively C Client 

. run 'make help' for more information, including a list of valid Build PRESETS. 
.    CONFIG:          [posix_fs-posix_platform-tls_bsp-senml-control_topic-memory_limiter-secure_file_transfer]
.    TARGET:          [osx-static-release]
.    COMPILER:        [cc] 
[cc] ####/xively-client-c/obj/osx/libxively/protobuf-generated/control_topic.pb-c.o
...
### omitted for brevity ###
...
In file included from ./src/import/tls/wolfssl/wolfssl/wolfcrypt/types.h:27:
./src/import/tls/wolfssl/wolfssl/wolfcrypt/settings.h:1573:14: error: 
      "For timing resistance / side-channel attack prevention consider using harden options" [-Werror,-W#warnings]
            #warning "For timing resistance / side-channel attack prevention consider using harden options"
             ^
1 error generated.
make: *** [/Users/westoncatron/heatworks/code/testing/xively-client-c/obj/osx/bsp/platform/posix/xi_bsp_rng_posix.o] Error 1
```

wcatron:
3.10.4 and 3.11.0
```
[...]

PR #179, opened 2018-01-16
Fix issue with enable debug throwing error.

wcatron:
Enable debug is set twice in the configuration file with enable debug enabled an error is thrown with it disbaled the error goes away #178 details the error.


atigyi:
Great catch @wcatron , thank you!

PR #183, opened 2018-02-06
Fix mbedTLS' memory_fs certificate parsing issues

Palantir555:
[ Description ]

This PR fixes the mbedTLS BSP. Currently it only works on POSIX, because it doesn't use `memory_fs` (hence, it doesn't use the hardcoded store). But certificate parsing would fail on the ESP32, or even POSIX when it's built with `memory_fs` instead of `posix_fs`.

- mbedTLS requires the root certificate buffer to be NULL terminated. This PR fixes the issue by updating our hardcoded buffer in `xi_RootCA_list.c` to include a NULL terminator.
- WolfSSL does NOT expect the buffer to be NULL-terminated, so fixing mbedTLS broke Wolf. I've updated the wolfSSL BSP to ignore null terminators in the certificate buffer (`buffer_len--`)
- `xi_RootCA_list.c` was generated using the `create_buffer.py` script. I've updated the script to append a NULL terminator to the string if necessary.
- `create_buffer.py` now uses the `XI_ROOTCA_LIST_BYTE_LENGTH` macro in the source file, instead of hardcoding the buffer length twice.

[ JIRA ]

[ Reviewer ]

[ QA ]
I've tested everything with 2 different hardcoded buffers: one null-terminated, one not:
```
POSIX/wolfSSL/memory_fs/with null terminator == OK
POSIX/mbedTLS/memory_fs/with null terminator == OK
POSIX/wolfSSL/memory_fs/without null terminator == OK
POSIX/mbedTLS/memory_fs/without null terminator == ERROR 19 (AS EXPECTED)

ESP32/wolfSSL/with null terminator == OK
ESP32/mbedTLS/with null terminator == OK
ESP32/wolfSSL/without null terminator == OK
ESP32/mbedTLS/without null terminator ==  ERROR 19 (AS EXPECTED)
```

[ Release Notes ]
Fix mbedTLS BSP certificate buffer parsing. It expected a NULL-terminated string.
The WolfSSL BSP has been updated to ignore the NULL terminator if present.


tools/create_buffer.py
```diff
@@ -10,20 +10,39 @@
...
 cpp_beg     = tabs_2_spaces( "#ifdef __cplusplus\nextern \"C\" {\n#endif" )
 cpp_end     = tabs_2_spaces( "#ifdef __cplusplus\n}\n#endif" )
 
+src_usage_msg = \
+"""\
+/* generated by create_buffer.py from a PEM file. Example:
+ * ./create_buffer.py --file_name res/trusted_RootCA_certs/xi_RootCA_list.pem
+ *    --array_name xi_RootCA_list --out_path src/libxively/tls/ --no-pretend
+ */
+"""
```
atigyi: 👍

tools/create_buffer.py
```diff
@@ -37,12 +56,17 @@ def load_file( file_name ):
...
+    else:
+        data.append( 0x00 )
+
+    return data
 
 def convert_to_c_array( data, array_name ):
-    out     = "const unsigned char %s[ %d ] = {" % ( array_name, len( data ) )
-    out    += NEWLINE
-    out    += INDENT
+    out  = "const unsigned char %s[ %s ] = {" % ( array_name, ROOTCA_LIST_LEN_MACRO )
```
atigyi: 👍

tools/create_buffer.py
```diff
@@ -37,12 +56,17 @@ def load_file( file_name ):
             data.append( struct.unpack( 'B', byte )[ 0 ] )
             byte = f.read( 1 )
 
-    return data[ 0 : -1 ]
+    if data[-1] in [0x0a, 0x00]: #['\n', '\0']
+        data[-1] = 0x00
```
atigyi: what about putting \n to the end instead of 0? Wouldn't \n keep wolfssl working without the last-char-ignore-function?

atigyi:
Seems ok, one minor question though. Otherwise: approved 💯

tools/create_buffer.py
Palantir555: It would. By default, there's already a `\n` at the end of this file's buffer. The previous version of the script was chopping it off, so I decided to get rid of it too.

That being said, don't you think it makes sense for this string to be null-terminated?

If the point is to avoid changes to the Wolf BSP, I'm happy to change it. There are a couple other small changes coming soon to that BSP, though.

atigyi: :) right, this is a string for sure. But putting 0 to the end differentiates it from the file version. So memory read will have cert string having 0 at the end, the file read will have \n. And the _prepare_ function will behave differently and we need two of them.
But JC, I leave this up to you since this isn't major issue, I think this is just a question of taste, so up to you, both versions solve the problem well!! Grats!

Palantir555: That's a very good point. I'm currently preparing another PR, but I'll update this to terminate with `\n`, and test again later.

Palantir555:
@atigyi I've implemented the change you suggested. The final result is definitely better, thanks again for the review :)

I had to remove the `const` qualifier from the certificate buffer, so that mbedTLS can modify the last character, but other than that, the changes are minor.

src/libxively/tls/certs/xi_RootCA_list.c
```diff
@@ -8,10 +8,15 @@
...
- * ./create_buffer.py --file_name res/trusted_RootCA_certs/xi_RootCA_list.pem \
- *    --array_name xi_RootCA_list --out_path src/libxively/tls/ --no-pretend */
-const unsigned char xi_RootCA_list[5941] = {
+#include "xi_RootCA_list.h"
+
+/* generated by create_buffer.py from a PEM file. Example:
+ * ./create_buffer.py --file_name res/trusted_RootCA_certs/xi_RootCA_list.pem
+ *    --array_name xi_RootCA_list --out_path src/libxively/tls/ --no-pretend
+ *
+ * The buffer is terminated with '\n' for consistency with POSIX filesystem reads
```
Palantir555: @atigyi I can remove this comment if you think it's too obvious. It wasn't clear for me at first sight.

atigyi: no, this is great here!

src/libxively/tls/certs/xi_RootCA_list.h
```diff
@@ -12,10 +13,10 @@ extern "C" {
 #endif
 
 #ifndef XI_ROOTCA_LIST_BYTE_LENGTH
-#define XI_ROOTCA_LIST_BYTE_LENGTH 5941
-#endif
+#define XI_ROOTCA_LIST_BYTE_LENGTH 5942
+#endif /* XI_ROOTCA_LIST_BYTE_LENGTH */
 
-extern const unsigned char xi_RootCA_list[XI_ROOTCA_LIST_BYTE_LENGTH];
+extern unsigned char xi_RootCA_list[ XI_ROOTCA_LIST_BYTE_LENGTH ];
```
atigyi: why not const?

atigyi: because mbedTLS BSP has to alter last character?

Palantir555: That's right. The application would crash with a `SIGBUS` error otherwise.

atigyi:
Approved. 💯 
(although I don's see where the .py file appends the `\n` to the buffer, but it certainly does)

Palantir555:
Thanks @atigyi !

There's no need to append `\n` to the buffer, because it already has one from the `file.read()`. The previous version would just chop it off with `return buf[ 0:-1 ]`, but using `return buf` saves that last character.

Merging this branch

PR #184, opened 2018-02-07
Support linking/including TLS libs outside xively-client-c

Palantir555:
[ Description ]

Until now, libxively would download, include and configure (via build options) whichever TLS library is selected in `XI_BSP_TLS`. That was OK because we were using our own versions of mbedTLS and WolfSSL, and we could decide the build configuration.

Now, the ESP32 SDK contains a pre-built, pre-configured and hardware-accelerated  mbedTLS, so we'd like to avoid downloading/compiling/linking our own version. There are 4 steps to this implementation:

1. Create a new build option that can be set for a given PRESET/XI_BSP_TLS combination
2. Avoid downloading the TLS library when such flag is enabled. Achieved by not setting TLS_LIB_PREPARE_CMD
3. Avoid building libxively's TLS BSP with incompatible build options by removing the inclusion of `make/mt-config/mt-tls-$(XI_BSP_TLS).mk`
3. Set your custom XI_TLS_LIB_INC_DIR path to the TLS library when the flag is set.

[ JIRA ]

[ Reviewer ]
@atigyi

[ QA ]
I'm using this patch to successfully build the ESP32 PRESET using the IDF SDK's mbedTLS.
It's automatically disabled when building for XI_BSP_TLS=wolfssl, and that works fine too.

[ Release Notes ]
Support linking and inclusion of TLS libraries outside the xively-client-c directory. This is part of the effort to get the ESP32 running its SDK's version of mbedTLS, as opposed of downloading and building our own.

### DIFF
--- a/make/mt-config/mt-presets.mk
+++ b/make/mt-config/mt-presets.mk
@@ -142,12 +142,14 @@ else ifeq ($(PRESET), ESP32)
     TARGET = $(TARGET_STATIC_REL)
     XI_BSP_PLATFORM = esp32
     XI_TARGET_PLATFORM = esp32
+    XI_USE_EXTERNAL_TLS_LIB = 1
 
 else ifeq ($(PRESET), ESP32_DEV)
     CONFIG = $(CONFIG_ESP32)
     TARGET = $(TARGET_STATIC_DEV)
     XI_BSP_PLATFORM = esp32
     XI_TARGET_PLATFORM = esp32
+    XI_USE_EXTERNAL_TLS_LIB = 1
 # -------------------------------------------------------
 # Fuzz Tests
 else ifeq ($(PRESET), FUZZ_TESTS)
--- a/make/mt-config/mt-presets.mk
+++ b/make/mt-config/mt-presets.mk
@@ -142,14 +142,18 @@ else ifeq ($(PRESET), ESP32)
     TARGET = $(TARGET_STATIC_REL)
     XI_BSP_PLATFORM = esp32
     XI_TARGET_PLATFORM = esp32
-    XI_USE_EXTERNAL_TLS_LIB = 1
+    ifeq ($(XI_BSP_TLS),mbedtls)
+        XI_USE_EXTERNAL_TLS_LIB = 1
+    endif
 
 else ifeq ($(PRESET), ESP32_DEV)
     CONFIG = $(CONFIG_ESP32)
     TARGET = $(TARGET_STATIC_DEV)
     XI_BSP_PLATFORM = esp32
     XI_TARGET_PLATFORM = esp32
-    XI_USE_EXTERNAL_TLS_LIB = 1
+    ifeq ($(XI_BSP_TLS),mbedtls)
+        XI_USE_EXTERNAL_TLS_LIB = 1
+    endif
 # -------------------------------------------------------
 # Fuzz Tests
 else ifeq ($(PRESET), FUZZ_TESTS)
--- a/make/mt-config/mt-presets.mk
+++ b/make/mt-config/mt-presets.mk
@@ -143,7 +143,7 @@ else ifeq ($(PRESET), ESP32)
     XI_BSP_PLATFORM = esp32
     XI_TARGET_PLATFORM = esp32
     ifeq ($(XI_BSP_TLS),mbedtls)
-        XI_USE_EXTERNAL_TLS_LIB = 1
+        XI_USE_EXTERNAL_TLS_LIB ?= 1
     endif
 
 else ifeq ($(PRESET), ESP32_DEV)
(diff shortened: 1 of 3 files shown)


Palantir555:
@DellaBitta @atigyi 

There must be a cleaner way of implementing this, but I didn't wanna spend too much time on it until I got your approval for the feature. Let me know what improvements you can think of.

make/mt-os/mt-esp32.mk
```diff
@@ -102,6 +102,14 @@ endif
 # mt-tls-mbedtls-esp32.mk. Or delete the line below to turn off TLS lib cross compilation.
 include make/mt-config/mt-tls-$(XI_BSP_TLS)-esp32.mk
 
+ifeq ($(XI_BSP_TLS),wolfssl)
+# WolfSSL TLS BSP configuration
+XI_CONFIG_FLAGS += -DXI_PROVIDE_WOLFSSL_SEED_GENERATOR
+XI_CONFIG_FLAGS += -DXI_PROVIDE_WOLFSSL_XTIME_XGMTIME
```
Palantir555: I git cherry-picked this code from my branch, and these config flags slipped through the cracks.

They'll make more sense when you see the rest of the ESP32 PR.

make/mt-config/mt-presets.mk
```diff
@@ -142,12 +142,18 @@ else ifeq ($(PRESET), ESP32)
     TARGET = $(TARGET_STATIC_REL)
     XI_BSP_PLATFORM = esp32
     XI_TARGET_PLATFORM = esp32
+    ifeq ($(XI_BSP_TLS),mbedtls)
+        XI_USE_EXTERNAL_TLS_LIB = 1
+    endif
```
DellaBitta: use ?= so we can override it on the command line if necessary.

make/mt-config/mt-presets.mk
```diff
@@ -142,12 +142,18 @@ else ifeq ($(PRESET), ESP32)
...
 
 else ifeq ($(PRESET), ESP32_DEV)
     CONFIG = $(CONFIG_ESP32)
     TARGET = $(TARGET_STATIC_DEV)
     XI_BSP_PLATFORM = esp32
     XI_TARGET_PLATFORM = esp32
+    ifeq ($(XI_BSP_TLS),mbedtls)
+        XI_USE_EXTERNAL_TLS_LIB = 1
```
DellaBitta: ditto

Palantir555:
@DellaBitta I've made `XI_USE_EXTERNAL_TLS_LIB` overridable, as you suggested. The ESP32 demo's build system (i.e. IDF's) won't handle `XI_USE_EXTERNAL_TLS_LIB=0` well, but it's a good starting point for users to fix it if they want to.

atigyi:
I guess the release notes are wrong, it corresponds to other PR: https://github.com/xively/xively-client-c/pull/183

make/mt-os/mt-esp32.mk
```diff
@@ -102,6 +102,14 @@ endif
...
+else ifeq ($(XI_BSP_TLS),mbedtls)
+XI_TLS_LIB_INC_DIR = $(IDF_PATH)/components/mbedtls/include
```
atigyi: so this means the default is to use the SDK's mbedTLS version. Is this ok?

Palantir555: The PRESET's default is still to use WolfSSL, but maybe I should change it. mbedTLS does have hardware acceleration in the ESP32, while wolf doesn't....

What do you think?

atigyi: what if I set the `XI_USE_EXTERNAL_TLS_LIB=0` in the make command line? What will happen? Which mbedTLS gets used? I think this line will be executed still which means the SDK's mbedTLS lib will be used. Won't it?

atigyi: So I see two value assignment to `XI_TLS_LIB_INC_DIR`. One in `mt-tls-mbedtls.mk` and one here `mt-esp32.mk`. In case `XI_USE_EXTERNAL_TLS_LIB == 0` both get executed. First the one in `mt-tls-mbedtls.mk` then this here in `mt-esp32.mk`, the second overrides the first. Which means setting NOT to use external TLS lib still assigns the external SDK's path.
Isn't this so?
But the final question is if the `XI_USE_EXTERNAL_TLS_LIB` flag works for both cases `1` and `0`?

atigyi:
One question regarding the actual functionality of the mbedTLS internal / external switch. Otherwise the feature is nice to have 👍

make/mt-os/mt-esp32.mk
Palantir555: Oh,  nice catch! 👍  You're right, that's broken. I can easily fix it with another `ifeq($(XI_USE_EXTERNAL_TLS_LIB),1)`, and I will.

`XI_USE_EXTERNAL_TLS_LIB=0` will still NOT work for the ESP32 either way, though. The example would still link the SDK's mbedTLS by default, so that would have to be solved, and things start getting more convoluted.

I can't think of a good reason to use our non-hardware-accelerated version instead of the SDK's, which is why `XI_USE_EXTERNAL_TLS_LIB` wasn't originally overridable from the command line.

Palantir555:
@atigyi Fix implemented. Thanks for the review :)

atigyi:
Approved 💯

PR #195, opened 2018-04-02
Fixes for case statement fallthroughs, which Debian gcc version 7.3.…

DellaBitta:
[ Description ]
I ran into some warnings thrown as errors when compiling on a fresh Ubuntu machine.  The warnings were about case statements automatically falling through. I've added break statements to fix these issue, tracing to code to ensure that the fallthrough wasn't intended in the first place.

[ Reviewer ]
@atigyi 

[ QA ]
Built running make and the mbedTLS implementation.  WolfSSL seems to have issues building on this development machine due to similar toolchain warnings-as-errors.

[ Release Notes ]
Fixes for building the xively-c-client with Debian GCC version 7.3.0 which produced some case statement fallthrough warnings.

### DIFF
--- a/src/import/cn-cbor/src/cn-get.c
+++ b/src/import/cn-cbor/src/cn-get.c
@@ -20,6 +20,7 @@ cn_cbor* cn_cbor_mapget_int(const cn_cbor* cb, int key) {
       if (cp->v.uint == (unsigned long)key) {
         return cp->next;
       }
+      break;
     case CN_CBOR_INT:
       if (cp->v.sint == (long)key) {
         return cp->next;
--- a/src/libxively/control_topic/xi_control_topic_layer.c
+++ b/src/libxively/control_topic/xi_control_topic_layer.c
@@ -196,6 +196,7 @@ xi_state_t xi_on_control_message( xi_context_handle_t in_context_handle,
                 xi_sft_on_connected( layer_data->sft_context );
 #endif
             }
+            break;
         }
         case XI_SUB_CALL_MESSAGE:
         {
@@ -212,6 +213,7 @@ xi_state_t xi_on_control_message( xi_context_handle_t in_context_handle,
 
             xi_sft_on_message( layer_data->sft_context, control_message );
 #endif
+            break;
         }
 
         default:;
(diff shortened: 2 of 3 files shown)


src/import/cn-cbor/src/cn-get.c
```diff
@@ -20,6 +20,7 @@ cn_cbor* cn_cbor_mapget_int(const cn_cbor* cb, int key) {
       if (cp->v.uint == (unsigned long)key) {
         return cp->next;
       }
+      break;
     case CN_CBOR_INT:
```
atigyi, line 24: this should be ok. BTW. this is a third party tool. Will be hard to track these changes. Also look at that https://github.com/cabo/cn-cbor/blob/master/src/cn-get.c they've already have this brake at place.

atigyi: Wow, Oct 11 2017: https://github.com/cabo/cn-cbor/commit/2e513d9d7a02e4ae74afa638d02d2b724b4fee92#diff-997e37b9907c2edf99c5eca675426940

atigyi: Hm, this means an update of the cn-cbor lib would solve this issue. And could introduce further ones. Let's keep this solution for now 👍

src/libxively/control_topic/xi_control_topic_layer.c
```diff
@@ -196,6 +196,7 @@ xi_state_t xi_on_control_message( xi_context_handle_t in_context_handle,
                 xi_sft_on_connected( layer_data->sft_context );
 #endif
             }
+            break;
         }
```
atigyi, line 200: wondering why lack of this break did not make trouble: a "fake" empty or trash message eaten up by the SFT module right after control topic subscription.

atigyi: I think because the cbor decode failed on litter, returned NULL `control_message` which is mapped to no-operation in the `xi_sft_on_message`.

atigyi:
Approved 👍

src/import/cn-cbor/src/cn-get.c
DellaBitta: Hah! nice!

src/libxively/control_topic/xi_control_topic_layer.c
DellaBitta: Yes, I had the same concern.  I was hoping that you would know. :)

PR #196, opened 2018-04-02
BACKOFF Compilation Error Fix

DellaBitta:
[ Description ]
Backoff LUT variables were named incorrectly in one of the not-often used predecessor configurations.

[ Reviewer ]
@atigyi 

[ QA ]
Compilation with Ubuntu/Debian GCC version 7.3 with the backoff_godmode configuration option.

[ Release Notes ]
Fixes for compilation in the BACKOFF module when it was configured for development mode.

### DIFF
--- a/src/libxively/xi_backoff_status_api.c
+++ b/src/libxively/xi_backoff_status_api.c
@@ -94,8 +94,8 @@ void xi_cancel_backoff_event()
 #ifdef XI_BACKOFF_GODMODE
 void xi_reset_backoff_penalty()
 {
-    xi_globals.backoff_status.backoff_lut_i = 0;
-    xi_globals.backoff_status.decay_lut_i   = 0;
+    xi_globals.backoff_status.backoff_lut = 0;
+    xi_globals.backoff_status.decay_lut   = 0;
 
     xi_cancel_backoff_event();
 }
--- a/src/libxively/xi_backoff_status_api.c
+++ b/src/libxively/xi_backoff_status_api.c
@@ -95,6 +95,7 @@ void xi_cancel_backoff_event()
 void xi_reset_backoff_penalty()
 {
     xi_globals.backoff_status.backoff_lut_i = 0;
+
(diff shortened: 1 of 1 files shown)


src/libxively/xi_backoff_status_api.c
```diff
@@ -94,8 +94,8 @@ void xi_cancel_backoff_event()
 #ifdef XI_BACKOFF_GODMODE
 void xi_reset_backoff_penalty()
 {
-    xi_globals.backoff_status.backoff_lut_i = 0;
-    xi_globals.backoff_status.decay_lut_i   = 0;
+    xi_globals.backoff_status.backoff_lut = 0;
+    xi_globals.backoff_status.decay_lut   = 0;
```
atigyi: hm. not sure about this. So here is the structure:
https://github.com/xively/xively-client-c/blob/master/src/libxively/xi_backoff_status.h
The `backoff_lut` and `decay_lut` are pointers on LUTs. The suffix `_i` is the current index on a certain LUT.
Doesn't the `xi_reset_backoff_penalty` function resets "only" the indices to 0? I don't think this reset function should just release the pointers on these tables. I guess the `_i` can stay for the backoff_lut. But yes, there is no `decay_lut_i`. Where it has gone, I don't know.

atigyi: Could you check where did the decay_lut_i go? Did it exist at all?

DellaBitta: Oh it gets even worse.  Nothing in our code even calls xi_reset_backoff_penalty

DellaBitta: Oh I see, it was meant for the application to be able to zero it out.  Interesting.  I'll look at it more.

DellaBitta: Good catch and my bad for not thinking things through.

atigyi:
Will be Approved as soon as Mr. Travis finishes without issue.

PR #197, opened 2018-04-02
Fix on KeepAlive Ping Write Errors

DellaBitta:
[ Description ]
Fix for handling socket disconnection after having write errors when attempting to send the keep alive ping request.
At disconnection the keepalive timeout guard task if forced to finish its job. Forced not from event scheduler but from the mqtt logic destructor. Although the task expected to be called only from event scheduler. This contract is guarded by the firing assert. Now timed event is cancelled before forcing its execution. 

[ JIRA ]
n/a

[ Reviewer ]
@atigyi

[ QA ]
Tested clients with backoff mode disabled. Clients used the same credentials and were disconnecting one another.  Set the MQTT keepalive time to the same time as backoff so that there were cases when the broker's duplicate credential disconnection event ran into the MQTT PingReq event (logged to ensure that this sometimes occurred).  Ran with the memory limiter enabled to ensure that memory wasn't leaking when these events occurred.  Ran a longevity test to see if repeated disconnections, and repeated MQTT PingReq teardowns, caused any instability.
Added 3 new keepalive integration test cases covering disconnecting socket after PINGREQ write error.

[ Release Notes ]
Bugfix - only in DEBUG mode - when socket disconnection happened before keepalive time interval passed after failing to write PINGREQ on the socket.
Fixes for when the MQTT Keep alive PingReq occurred at the same tick as socket was closed.

### DIFF
--- a/src/libxively/mqtt/logic/xi_mqtt_logic_layer.c
+++ b/src/libxively/mqtt/logic/xi_mqtt_logic_layer.c
@@ -605,6 +605,11 @@ xi_mqtt_logic_layer_close_externally( void* context, void* data, xi_state_t in_o
     /* let's stop the current task */
     if ( layer_data->current_q0_task != 0 )
     {
+        if ( NULL != layer_data->current_q0_task->timeout.ptr_to_position )
+        {
+            xi_evtd_cancel( event_dispatcher, &layer_data->current_q0_task->timeout );
+        }
+
         layer_data->current_q0_task->logic.handlers.h4.a3 = XI_STATE_TIMEOUT;
         xi_evtd_execute_handle( &layer_data->current_q0_task->logic );
         xi_mqtt_logic_free_task( &layer_data->current_q0_task );
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,3 +100,4 @@ script:
         fi;
     fi;"
 
+
--- a/doc/doxygen/build.sh
+++ b/doc/doxygen/build.sh
@@ -5,3 +5,4 @@ cd ../api
 rm -rf output
 doxygen doxygen.config
 cd ..
+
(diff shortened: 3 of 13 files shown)


atigyi:
The case this change targets is when the library failed to send the MQTT KEEPALIVE message. The previous handling of this situation was to wait the keepalive timeout and then do a libxively internal reconnect. Although this reconnect probably did not happen because broker closed the connection before in .5 * keepalive secs. Right?!
What was the problem with this previous "error handling" mechanism? The assert: 
`assert( NULL == task->timeout.ptr_to_position ); `?

What this version has as a difference is that the `on_keepalive_timeout_expiry` event is not scheduled. So it won't hit back when connection closes. Why does the connection close here? Because server closes it. Why? Keepalive timeout or same credential reuse?

"This would put the library in a confusing state." I would rather explore what this "confusing state" is and try to fix it than dodge it. Why? Because this confusing state might be reached on a different path too. Knowing this state would help if there are other paths to reach this state.

We should reproduce this issue with integration test case(s) and fix it that way.

src/libxively/mqtt/logic/xi_mqtt_logic_layer.c
```diff
@@ -605,6 +605,11 @@ xi_mqtt_logic_layer_close_externally( void* context, void* data, xi_state_t in_o
     /* let's stop the current task */
     if ( layer_data->current_q0_task != 0 )
     {
+        if ( NULL != layer_data->current_q0_task->timeout.ptr_to_position )
```
atigyi, line 608: This is the actual bugfix. The rationale behind:
This part of code is the close_externally, which is the destruction of the layerchain. Ongoing task is executed at destruction, to notify the application callbacks about task progress (PUBLISH / SUBSCRIBE failure). But in this case the task is a timed task which sits inside the event dispatcher's timed event's queue.
The assert in question is checking the idea behind: keepalive task should be executed by event scheduler which actually resets the timed event. But here not the event scheduler executes the task but the mqtt logic layer's destructor function. So here we cancel the timed event before execution which will silence the assert and the idea / contract guarded by the asserts kept valid.

PR #199, opened 2018-04-06
Feature/wolfssl implicitfallthrough

DellaBitta:
[ Description ]
Disables the implicit fall-through warnings when building wolfSSL. In the future we can re-enable them when we upgrade wolf versions (which has since added explicit fall-through instrumentation).  For now, though, the latest wolf Versions don't seem to play well with the ESP32 build, and there doesn't seem to be much need to upgrade it at this time.

[ Reviewer ]
@atigyi 

[ QA ]
Build and tested POSIX library and examples.
Built ESP32 build.

[ Release Notes ]
Build configuration change for building WolfSSL version on Ubuntu.

### DIFF
--- a/src/import/cn-cbor/src/cn-get.c
+++ b/src/import/cn-cbor/src/cn-get.c
@@ -20,6 +20,7 @@ cn_cbor* cn_cbor_mapget_int(const cn_cbor* cb, int key) {
       if (cp->v.uint == (unsigned long)key) {
         return cp->next;
       }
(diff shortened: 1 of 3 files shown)


atigyi:
So this makes ESP32 compatible "old" wolfSSL compile with newest linux compiler?

atigyi:
Approved.

DellaBitta:
Yes, that's corrrect.

PR #200, opened 2018-04-27
Bugfix/xi_desc_alloc_assert

DellaBitta:
[ Description ]
Fix in debug builds for a potential assertion scenario in our data model.  The scenario would occur when the library needed to allocate a dynamically sized buffer along with the xi_data_desc_t structure to track this allocation but didn't have enough free space to allocate both.  

This issue does not have a behavioral impact in release mode.

The problem would arise when the first allocation, the xi_data_desc_t, succeeds but there wasn't enough memory to subsequently create the dynamically-sized buffer.  Generally both allocations take place, and then the structure of the xi_data_desc_t is initialized afterwards with the buffer size data and a pointer to the buffer.  And additional data ownership field tracks who should deallocate the buffer later when the xi_data_desc_t structure is freed.  This last field is important to xi_free_desc() as it needs to know if it should free the buffer that the xi_data_desc_t points to, or if someone else owns that responsibility.

In the case when the xi_data_desc_t structure was properly allocated but the subsequent dynamically allocated buffer failed in allocation, the xi_data_desc_t initialization code was skipped.  When the de-allocation of the xi_data_desc_t structure occurs immediately following to cleanup the state after the error, the function xi_free_desc() would throw an assertion because the data ownership field wasn't set.

This assertion scenario has been fixed by setting up the required fields in the xi_data_desc_t before the dynamically allocated buffer malloc attempt occurs. This way if the malloc of the buffer fails, then xi_free_desc() has the information that it needs to operate on the xi_data_desc_t.

[ Reviewer ]
@atigyi 

[ QA ]
Numerous Unit Tests were created to test the various data buffer creation functions in xi_data_desc.c. These new unit tests exercise the scenarios when there's not enough memory to allocate the xi_data_desc_t, and another test to if it can allocate the xi_data_desc_t tracking structure, but not the related data buffer.

[ Release Notes ]
Fixes for assertions which would occur in xi_data_desc_t creation functions when the runtime doesn't have enough free memory to allocate the buffer.  This was only a debug assertion and this change has no impact on the behavior of the library in release mode.

### DIFF
--- a/src/tests/utests/xi_utest_data_desc.c
+++ b/src/tests/utests/xi_utest_data_desc.c
@@ -244,6 +244,263 @@ XI_TT_TESTCASE_WITH_SETUP(
         tt_fail();
     } )
 
+#ifndef XI_MEMORY_LIMITER_ENABLED
+#define XI_MEMORY_LIMITER_ENABLED 1
+#endif 
+
+#ifdef XI_MEMORY_LIMITER_ENABLED
+XI_TT_TESTCASE(
+    utest__xi_make_desc_from_string_copy__valid_data__not_enough_memory_for_buffer,
+    {
+        /* size values */
+        size_t num_characters     = 700;
+        xi_state_t ret_state      = XI_STATE_OK;
+        char* origin_string       = NULL;
+        xi_data_desc_t* data_desc = NULL;
+        
+        const size_t allocation_space   = num_characters * sizeof( char ) +
+                                          sizeof( xi_data_desc_t ) +
+                                          sizeof( xi_data_desc_t ) +
+                                          2*sizeof( xi_memory_limiter_entry_t );
+        const size_t allocation_footprint = num_characters* sizeof( char ) +
+                                            sizeof( xi_memory_limiter_entry_t );
+        const size_t memory_limit_size = XI_MEMORY_LIMITER_SYSTEM_MEMORY_LIMIT +
+                                            allocation_space;
+
+        /* assumptions */
+        tt_assert( XI_MEMORY_LIMITER_SYSTEM_MEMORY_LIMIT > allocation_footprint );
+
+        xi_state_t memory_set_result = xi_set_maximum_heap_usage( memory_limit_size );
+        tt_assert( memory_set_result == XI_STATE_OK );
+
+        XI_ALLOC_BUFFER_AT( char, origin_string, num_characters, ret_state );
+        size_t i = 0;
+        for ( ; i < num_characters - 1; ++i )
+        {
+            origin_string[i] = 'a';
+        }
+        origin_string[num_characters-1] = '\0';
+        tt_int_op( strlen( origin_string ), ==, num_characters - 1 );
+
+        /* test behavior */
+        data_desc = xi_make_desc_from_string_copy( origin_string );
+        tt_ptr_op( data_desc, ==, NULL );
+        
+        goto end;
+
+    err_handling:
+        tt_fail();
+    end:
+        if( data_desc != NULL )
+        {
+            xi_free_desc( &data_desc );
+        }
+        XI_SAFE_FREE( origin_string );
+        xi_memory_limiter_set_limit( XI_MEMORY_LIMITER_SYSTEM_MEMORY_LIMIT +
+                                     XI_MEMORY_LIMITER_APPLICATION_MEMORY_LIMIT );
+
+    } )
+
+    XI_TT_TESTCASE(
+    utest__xi_make_desc_from_string_copy__valid_data__not_enough_memory_for_descriptor,
+    {
+        /* size values */
+        size_t num_characters     = 700;
+        xi_state_t ret_state      = XI_STATE_OK;
+        char* origin_string       = NULL;
+        xi_data_desc_t* data_desc = NULL;
+        
+        const size_t allocation_space   = num_characters * sizeof( char ) +
+                                          sizeof( xi_data_desc_t ) +
+                                          sizeof( xi_memory_limiter_entry_t );
+        const size_t allocation_footprint = num_characters* sizeof( char ) +
+                                            sizeof( xi_memory_limiter_entry_t );
+        const size_t memory_limit_size = XI_MEMORY_LIMITER_SYSTEM_MEMORY_LIMIT +
+                                            allocation_space;
+
+        /* assumptions */
+        tt_assert( XI_MEMORY_LIMITER_SYSTEM_MEMORY_LIMIT > allocation_footprint );
+
+        xi_state_t memory_set_result = xi_set_maximum_heap_usage( memory_limit_size );
+        tt_assert( memory_set_result == XI_STATE_OK );
+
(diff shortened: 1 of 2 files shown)


atigyi:
The solution looks good. The unit tests seem overlapping and too fragile at first sight. Please give me time to come up with a more maintenance friendly unit test version.

src/tests/utests/xi_utest_data_desc.c
```diff
@@ -244,6 +244,259 @@ XI_TT_TESTCASE_WITH_SETUP(
...
+                                          sizeof( xi_data_desc_t ) +
+                                          sizeof( xi_data_desc_t ) +
+                                          2*sizeof( xi_memory_limiter_entry_t );
+        const size_t allocation_footprint = num_characters* sizeof( char ) +
+                                            sizeof( xi_memory_limiter_entry_t );
+        const size_t memory_limit_size = XI_MEMORY_LIMITER_SYSTEM_MEMORY_LIMIT +
+                                            allocation_space;
+
+        /* assumptions */
+        tt_assert( XI_MEMORY_LIMITER_SYSTEM_MEMORY_LIMIT > allocation_footprint );
```
atigyi: what does this assert mean? System memory limit must exceed allocation footprint.
System memory limit is the minimal amount of memory the libxively needs for minimal functionality.
Allocation footprint is a memory limiter entry + the test string size.
How does this two come together with this relation?

src/tests/utests/xi_utest_data_desc.c
```diff
@@ -244,6 +244,259 @@ XI_TT_TESTCASE_WITH_SETUP(
...
+        /* size values */
+        size_t num_characters     = 700;
+        xi_state_t ret_state      = XI_STATE_OK;
+        char* origin_string       = NULL;
+        xi_data_desc_t* data_desc = NULL;
+        
+        const size_t allocation_space   = num_characters * sizeof( char ) +
+                                          sizeof( xi_data_desc_t ) +
+                                          sizeof( xi_data_desc_t ) +
+                                          2*sizeof( xi_memory_limiter_entry_t );
```
atigyi: for first and second sight this looks magic. What are the components `allocation_space` is built up?

src/tests/utests/xi_utest_data_desc.c
```diff
@@ -244,6 +244,259 @@ XI_TT_TESTCASE_WITH_SETUP(
         tt_fail();
     } )
 
+#ifdef XI_MEMORY_LIMITER_ENABLED
+XI_TT_TESTCASE(
+    utest__xi_make_desc_from_string_copy__valid_data__not_enough_memory_for_buffer,
+    {
+        /* size values */
+        size_t num_characters     = 700;
```
atigyi: Is it a rule to define variables at the beginning? Isn't it C89 rule?

src/tests/utests/xi_utest_data_desc.c
```diff
@@ -244,6 +244,259 @@ XI_TT_TESTCASE_WITH_SETUP(
...
+    end:
+        if( data_desc != NULL )
+        {
+            xi_free_desc( &data_desc );
+        }
+        XI_SAFE_FREE( origin_buffer );
+        xi_memory_limiter_set_limit( XI_MEMORY_LIMITER_SYSTEM_MEMORY_LIMIT +
+                                     XI_MEMORY_LIMITER_APPLICATION_MEMORY_LIMIT );
+
+    } )
```
atigyi:
```
441     XI_TT_TESTCASE(
442     utest__xi_make_desc_from_buffer_copy__valid_data__not_enough_memory_for_descriptor_atigyi,
443     {
444         /* size values */
445         const unsigned char origin_buffer[] = "Test data";
446         const size_t buffer_size_in_bytes = sizeof(origin_buffer);
447 
448         const size_t allocation_space   = sizeof( xi_data_desc_t ) +
449                                           sizeof( xi_memory_limiter_entry_t ) +
450                                           buffer_size_in_bytes +
451                                           sizeof( xi_memory_limiter_entry_t ) - 1;
452 
453         const size_t memory_limit_size = XI_MEMORY_LIMITER_SYSTEM_MEMORY_LIMIT +
454                                          allocation_space;
455 
456         xi_state_t memory_set_result = xi_set_maximum_heap_usage( memory_limit_size );
457         tt_assert( memory_set_result == XI_STATE_OK );
458 
459         /* test behavior */
460         xi_data_desc_t* data_desc = xi_make_desc_from_buffer_copy( origin_buffer, buffer_size_in_bytes );
461         tt_ptr_op( data_desc, ==, NULL );
462 
463     end:
464         xi_free_desc( &data_desc );
465         xi_memory_limiter_set_limit( XI_MEMORY_LIMITER_SYSTEM_MEMORY_LIMIT +
466                                      XI_MEMORY_LIMITER_APPLICATION_MEMORY_LIMIT );
467 
468     } )
```

atigyi: I would go with the implementation above. What it has in difference are:
- it uses an initialized array on the stack instead of using XI_ALLOC_BUFFER. This simplifies the test cases a bit by not needing the err_handling and simpler, bit clearer calculation of `allocation_space`.
- also I've removed the footprint check since I don't get what it is for. Maybe you will describe it for me.
- no need for guarding `xi_free_desc` it must handle NULL values.

atigyi: Let's discuss these suggestions and after please update all test cases to make them simpler.
Also could you please add a comment how the allocation_space is calculated.

src/tests/utests/xi_utest_data_desc.c
```diff
@@ -244,6 +244,259 @@ XI_TT_TESTCASE_WITH_SETUP(
         tt_fail();
     } )
 
+#ifdef XI_MEMORY_LIMITER_ENABLED
```
atigyi, line 247: I've checked this condition. Unfortunately I did not find better solution. The only tool in our hands to drive these `*make_desc*` functions to the test scenario is the memory limiter. But at least we have this tool. And congrats great trick with this tool to use it for this purpose.

DellaBitta: Yes, this doesn't feel great to me, either, but ... here we are. :)

src/tests/utests/xi_utest_data_desc.c
DellaBitta: C89 / ANSI.

src/tests/utests/xi_utest_data_desc.c
DellaBitta: OK, sure!  I can create a variable for each one individually.

src/tests/utests/xi_utest_data_desc.c
DellaBitta: If this isn't true then there wouldn't be enough memory to support the test itself since the allocations of the test will not come from the system memory limit.  

This was in the other memory limiter tests so I kept in here to be safe.

DellaBitta: I can add a comment to what it's intention is.

